PR #9, opened 2017-06-29
view and insert rule example for postgresql

dangowrt:
complete default_packages view and introduce ON INSERT rule.
in mysql this will have to be done using [`CREATE TRIGGER`](https://dev.mysql.com/doc/refman/5.7/en/trigger-syntax.html)


aparcar:
thanks you very much!

PR #21, opened 2017-07-08
update server package dependencies

dangowrt:
dependency on python3-distutils was masked by inaccurate packaging of
python-yaml. Now that python-yaml is packaged more tightly it blew up.


aparcar:
thanks again!

PR #99, opened 2018-06-10
Ansible cleanup and commenting.

oneirosFade:
After a brief chat with Igel I've gone over the Ansible and done a (admittedly opinionated) cleanup, fixed some issues, made it more (hopefully) robust/resilient. There are a couple outstanding issues, noted with @TODO in the code.

I've also (since host_vars are important now) added three dummy host_vars files to demonstrate their use in the testing environment.


aparcar:
Hi @oneirosFade thank you very much for your work. Could you fix the todos? Then I'm happy to squash and merge this PR!

In parallel there were efforts to simplify the setup by using Psycogs, maybe we should fix https://github.com/aparcar/attendedsysupgrade-server/pull/97 first and then make Ansible use Psycogs as well?

oneirosFade:
I've cleaned up the TODO items as best I can, and it should be ready for merge now.

As I understand it, Ansible is using Psycogs for pgsql actions (hence why it's installed on the targets).

That being said, the cli.py doesn't execute, citing an error that the 'releases' relation doesn't exist. I'm not familiar with the software itself, so I don't know that I can help here. I don't think this is an issue with Vagrant/Ansible, and will open an issue against the repo if you agree.

aparcar:
> will open an issue against the repo if you agree.

Sure please go ahead. I'll review the PR now

PR #119, opened 2018-08-15
Fast update

aparcar:
enable multi thread builders and updaters
this is nice as the package update is quite slow, now we can 8 in parallel


aparcar:
Thanks for the information, somehow the stuff is called `threading` in Python, so I end up with that. Generally I should write somewhat more informative commit messages :P

PR #120, opened 2018-08-15
server as a single package

aparcar:
startable via `python3 server`
also let the server start the updater, boss and garbagecollector


aparcar:
currently testable at chef.apracar.org

PR #162, opened 2019-03-06
supported status based on metadata

aparcar:
the metadata of recent images may contain signatures via usign/ucert.
these are used to trust the server and active workers


aparcar:
Close due to Redis rewrite.

PR #168, opened 2019-03-15
rework: less reloading of files

aparcar:
unify all config settings on first load


aparcar:
Already merged in `ucert` branch

PR #242, opened 2020-03-22
use rq-scheduler for continuous upstream sync

aparcar:


aparcar:
Replaced via https://github.com/aparcar/asu/commit/79e6bb6f6188ef0914cf16dde671d8cf14d2f500 The RQ scheduler would use RQ workers to update things, however long term multiple workers on different servers should build images while not having access to the servers `/json` folder.

PR #253, opened 2020-06-15
api: add latest call

aparcar:
The call is easier to process on the client side  and can later be
hosted as static files.


aparcar:
Superseded.

PR #265, opened 2021-03-14
api: add `version_code` and `packages_versions`

aparcar:
Both functions are described in the README.

### DIFF
--- a/asu/api.py
+++ b/asu/api.py
@@ -74,6 +74,48 @@ def get_queue() -> Queue:
     return g.queue
 
 
+def validate_packages(req):
+    if "packages" in req and "packages_versions" in req:
+        return (
+            {
+                "status": "duplicate_package_list",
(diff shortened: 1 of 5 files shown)


asu/api.py
```diff
@@ -74,6 +74,48 @@ def get_queue() -> Queue:
...
+        )
+    elif "packages" in req:
+        pass
+    elif "packages_versions" in req:
+        req["packages"] = req["packages_versions"].keys()
+    else:
+        # no packages to check
+        return
+
+    req["packages"] = set(req["packages"]) - {"kernel", "libc", "libgcc"}
```
dangowrt, line 94: Shouldn't this contain `libpthread` as well? I filter those already in `auc`.

aparcar: I never got a error here but sure can add that too

PR #269, opened 2021-03-19
Add ucert support

aparcar:
This allows to use define a CA_PUBKEY which is added to running devices. As a consequence a secret upgrade path is possible because OpenWrts `sysupgrade` can check ucert signatures.


aparcar:
@dangowrt here we go

PR #294, opened 2021-08-23
added build context to the docker-compose file because docker-compose…

egon0:
… complains about it


aparcar:
Thanks, I'm using podman-compose which did not complain but I'm happy if it works for both now.

PR #337, opened 2022-03-15
build: disable ext4 file creation

aparcar:
This step disables the creation of firmware with the ext4 filsystem.
For OpenWrt it is not possible to upgrade a device running on an ext4
filesystem without losing modified files. For SquashFS a filesystem
overlay is added which is recovered after the sysupgrade.


aparcar:
@dangowrt thoughts?

dangowrt:
I don't think ext4 should be disabled unconditionally. 
OpenWrt **does** support keeping configuration files when sysupgrade'ing also for ext4, jffs2 and ubifs formatted read-write rootfs. Other features (like keeping **all** modified files or **factory reset**) are unavailable when not using squashfs, as they depend on using a read-only rootfs with read-write overlay on top.
Hence it would be better to include the filesystem type in the image request API (optional, to keep compatible with existing clients) and if given, build only the filesystem type requested.

aparcar:
@dangowrt I chose a different path for the implementation, it individually disables non used params. What do you think?

asu/build.py
```diff
@@ -123,6 +123,26 @@ def setup_ib():
...
+            config_path = cache / subtarget / ".config"
+            config = config_path.read_text()
+
+            if req["filesystem"] == "squashfs":
+                config = config.replace(
+                    "CONFIG_TARGET_ROOTFS_EXT4FS=y",
+                    "# CONFIG_TARGET_ROOTFS_EXT4FS is not set",
+                )
+
+            elif req["filesystem"] == "ext4":
```
dangowrt, line 137: Please also support `ubifs` as well as `jffs2`, as then all 4 filesystem types used as rootfs within OpenWrt are covered.

asu/build.py
```diff
@@ -123,6 +123,26 @@ def setup_ib():
...
 
+        if req.get("filesystem"):
+            log.info("Limit filesytem to %s", req["filesystem"])
+            config_path = cache / subtarget / ".config"
+            config = config_path.read_text()
+
+            if req["filesystem"] == "squashfs":
+                config = config.replace(
+                    "CONFIG_TARGET_ROOTFS_EXT4FS=y",
+                    "# CONFIG_TARGET_ROOTFS_EXT4FS is not set",
```
dangowrt, line 134: I guess here you should also disable JFFS2 and UBIFS.

PR #338, opened 2022-03-16
build: reset active filesystem after each build

aparcar:
The active filesystem was wrongly only set on initial imagebuilder setup
instead of every time an image is build. This was due to the initial
implementation which would disable ext4 in general instead of disabling
filesystems specifically per request.

Now also support JFFS2 and UBIFS filesystems.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -4,7 +4,7 @@ import re
 import subprocess
 from datetime import datetime
 from pathlib import Path
-from shutil import rmtree
+from shutil import rmtree, copyfile
 
 import requests
 from rq import get_current_job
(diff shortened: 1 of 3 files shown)


asu/build.py
```diff
@@ -282,6 +262,27 @@ def download_file(filename: str, dest: str = None):
 
     log.debug("Created store path: %s", req["store_path"] / bin_dir)
 
+    config_path = cache / subtarget / ".config"
+    config = config_path.read_text()
+
+    # enables all filesystem if none is specified
+    for filesystem in ["squashfs", "ext4fs", "ubifs", "jffs2"]:
+        # this implementation uses `startswith` since a running device thinks
+        # it's running `ext4` or `ubi` while it's really called `ext4fs`
```
dangowrt, line 271: This is true for `ext4` which becomes `ext4fs`. However, `ubifs` is just called `ubifs` always (the backing device for UBIFS is always a UBI *volume* which lives on a UBI device which typically lives on a MTD partition of a NAND MTD device). So using `startswith` is of course ok, just the comment needs to be more clear about the reason being `ext4fs` / `ext4` particularly and *not* some generic rule for which `ubifs` / `ubi` would be another example (because as stated above, it is not. The filesystem is also called `ubifs` by the Linux kernel.)

Sidenode: The `-ubi` in the image name of some devices does **not** indicate that they are using UBIFS as rootfs. Production systems using UBI to manage NAND storage will just squashfs, just like you'd do on x86.

asu/build.py
```diff
@@ -282,6 +262,27 @@ def download_file(filename: str, dest: str = None):
 
     log.debug("Created store path: %s", req["store_path"] / bin_dir)
 
+    config_path = cache / subtarget / ".config"
+    config = config_path.read_text()
+
+    # enables all filesystem if none is specified
+    for filesystem in ["squashfs", "ext4fs", "ubifs", "jffs2"]:
```
dangowrt, line 269: Shouldn't this all still just happen under condition
```
if req.get("filesystem"):
```
?

dangowrt (approved):
Apart from a misleading comment this looks good to me (and I'm not into Python...)

PR #373, opened 2022-07-27
Update Python version to 3.10 and use `slim` image

### LINKED ISSUE #372: Janitor uses Python 3.8 features but Docker images aren't updated past 3.7
https://github.com/openwrt/asu/commit/f4df611730c085b53d3ab4f9ff26d07fa0732af4 and https://github.com/openwrt/asu/commit/0ae4fb361888215f0cbe3416b05da710aba75a9d causes the default Docker config to fail.

magneticflux-:
Closes #372 

Signed-off-by: Mitchell Skaggs <skaggsm333@gmail.com>

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -1,4 +1,4 @@
-FROM python:3.7-buster
+FROM python:3.10-slim
 
 RUN useradd -c "OpenWrt Builder" -m -d /home/build -s /bin/bash build
 


aparcar:
Thanks, problem is that current worker images are using the official `openwrt/imagebuilder` container which uses an old Python version. Using a newer Python version comes with a different Pickle version which is used to communicate by `rq`.

I think there is an option for `rq` to explicitly use the "old" pickle format, could you please check that out?

```
build@7074cb0a47da:~/openwrt$ python3
Python 3.7.3 (default, Jan 22 2021, 20:04:44) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
```

magneticflux-:
Unfortunately there is no such option for clients of RQ without patching it ourselves: https://github.com/rq/rq/issues/1168

I could make a PR on Gitlab here to update from Buster to Bullseye (Debian 10 to 11): https://gitlab.com/openwrt/docker/-/blob/master/Dockerfile.base#L1. Buster's default Python3 is 3.7, but Bullseye's default is 3.9. However, this also updates GCC from 8.3 to 10.2.

aparcar:
I created a PR to upgrade the ImageBuilders, let's see if it works. If that's not an option we could also switch to the JSON serializer, however I'm not sure if that slows down things.

aparcar:
I updated the imagebuilder container, this should work now.

PR #374, opened 2022-07-28
api: allow clients to send their identification

aparcar:
This allows to track the clients used to request updates. Ideally this help to
identify what versions are mostly used and drop support for outdated version.

A client can now send the "client" field in the format "name - version".


aparcar:
@dangowrt could you please implement this for `auc`?

PR #402, opened 2022-10-04
api: drop server side handling of package_changes

aparcar:
This causes more issues than good. The server must only provide such information and the client should request the user what do to or may "auto default" to what `package_changes` suggest.

Changing this on the server side confuses clients that compare the requested list of packages with the resulting list of packages.


aparcar:
CC: @dangowrt I'll implement this for LuCI, please do so for `auc`

PR #404, opened 2022-10-09
branches: add revision and mandatory to changes

aparcar:
The revision can be used to "date" package changes so that updates between i.e. snapshots won't be suggested every time.

The mandatory field is used to specify if a changes is selectable or not. If a package is dropped, the changes will always be selected.

Signed-off-by: Paul Spooren <paul.spooren@rhebo.com>

### DIFF
--- a/asu/branches.yml
+++ b/asu/branches.yml
@@ -22,8 +22,12 @@ branches:
       - 22.03.0-rc6
       - 22.03-SNAPSHOT
     package_changes:
-      kmod-nft-nat6:
-      firewall: firewall4
+      - source: kmod-nft-nat6
+        revision: 19160
+        mandatory: true
+      - source: firewall
+        target: firewall4
+        revision: 18611
 
   "21.02":
     enabled: true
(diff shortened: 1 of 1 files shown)


dangowrt:
> If a package is dropped, the changes will always be selected.

A package which should be dropped **optionally** is also an imaginable situation, let's say e.g. in future we decide to ditch uclient-fetch in favor of yet another wget implenentation we would also drop libustream-wolfssl from the default package set. In that case, the package is dropped, but the user should have the option to keep it (as it is still present as a package in the tree, it's just no longer part of the default set and has been replaced there).

asu/branches.yml
```diff
@@ -67,5 +71,9 @@ branches:
...
       - SNAPSHOT
     package_changes:
-      kmod-nft-nat6:
-      firewall: firewall4
+      - source: kmod-nft-nat6
+        revision: 20282
+        mandatory: true
+      - source: firewall
+        target: firewall4
+        revision: 18611
```
dangowrt, line 79: As both changes (firewall -> firewall4, kmod-nft-nat6 -> null) are present already in the 22.03 release branch, why do you list them again in SNAPSHOT? That makes it much harder for the client to decide which changes to include...

PR #416, opened 2022-10-30
api: add rootfs_size_mb parameter

### LINKED ISSUE #415: [generic images] Allow setting of CONFIG_TARGET_ROOTFS_PARTSIZE
It would be great if the client could request a custom CONFIG_TARGET_ROOTFS_PARTSIZE so that there would be no need to resize after flashing (or so in case 100MB is somehow not enough for the packages requested).

rany2:
The parameter allows the client to specify a custom value for
the CONFIG_TARGET_ROOTFS_PARTSIZE. This is especially useful
for the generic X86 images.

Fixes #415

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -308,6 +308,8 @@ def build(req: dict):
         f"EXTRA_IMAGE_NAME={packages_hash}",
         f"BIN_DIR={req['store_path'] / bin_dir}",
     ]
+    if rootfs_size_mb := req.get("rootfs_size_mb"):
+        build_cmd.append(f"ROOTFS_PARTSIZE={rootfs_size_mb}")
 
     log.debug("Build command: %s", build_cmd)
 
--- a/asu/openapi.yml
+++ b/asu/openapi.yml
@@ -265,6 +265,15 @@ components:
             Ability to specify filesystem running on device. Attaching this
             optional parameter will limit the ImageBuilder to only build
             firmware with that filesystem.
+        rootfs_size_mb:
+          type: integer
+          minimum: 1
+          maximum: {{ rootfs_size_mb_max }}
+          example: 100
+          description: |
+            Ability to specify a custom CONFIG_TARGET_ROOTFS_PARTSIZE for the
+            resulting image. Attaching this optional parameter will cause
+            ImageBuilder to build a rootfs with that size in MB.
         defaults:
           type: string
           description: |
--- a/misc/config.py
+++ b/misc/config.py
@@ -16,6 +16,9 @@ CACHE_PATH = None
 # where to store JSON files
 JSON_PATH = Path.cwd() / "public/json/v1/"
 
+# maximum custom rootfs size
+MAX_CUSTOM_ROOTFS_SIZE_MB = 100
+
 # manual mapping of package ABI changes
 MAPPING_ABI = {"libubus20191227": "libubus"}
 
--- a/asu/asu.py
+++ b/asu/asu.py
@@ -39,6 +39,7 @@ def create_app(test_config: dict = None) -> Flask:
         ALLOW_DEFAULTS=False,
         ASYNC_QUEUE=True,
         BRANCHES_FILE=getenv("BRANCHES_FILE"),
+        MAX_CUSTOM_ROOTFS_SIZE_MB=100,
     )
 
     if not test_config:
@@ -128,6 +129,12 @@ def create_app(test_config: dict = None) -> Flask:
         if not app.config["REDIS_CONN"].hexists("mapping-abi", package):
             app.config["REDIS_CONN"].hset("mapping-abi", package, source)
 
-    cnxn.add_api("openapi.yml", validate_responses=app.config["TESTING"])
+    cnxn.add_api(
+        "openapi.yml",
+        arguments={
+            "rootfs_size_mb_max": app.config["MAX_CUSTOM_ROOTFS_SIZE_MB"],
+        },
+        validate_responses=app.config["TESTING"],
+    )
 
     return app
--- a/asu/common.py
+++ b/asu/common.py
@@ -111,6 +111,7 @@ def get_request_hash(req: dict) -> str:
                 str(req.get("diff_packages", False)),
                 req.get("filesystem", ""),
                 get_str_hash(req.get("defaults", "")),
+                str(req.get("rootfs_size_mb", "")),
             ]
         ),
         32,


misc/config.py
```diff
@@ -16,4 +16,7 @@
 # where to store JSON files
 JSON_PATH = Path.cwd() / "public/json/v1/"
 
+# maximum custom rootfs size
+MAX_CUSTOM_ROOTFS_SIZE = 14748
```
aparcar, line 20: Do you mind adding the unit to the variable name? Like MB?

asu/build.py
```diff
@@ -300,6 +301,21 @@ def download_file(filename: str, dest: str = None):
             cache_workdir / ".config",
         )
 
+    # Prevent DoS by large int<->str conversions (https://github.com/python/cpython/issues/95778)
+    # This sets a hard limit of < 1TB as the max processible rootfssize (with the actual limit set in config)
+    if req.get("rootfssize") and len(req.get("rootfssize")) < 7:
```
aparcar, line 306: I think this requires more logic now, the rootfssize should be reset after (or before) each build.

rany2:
I've pushed a new commit. Could you please check if I've addressed your concerns properly?

rany2:
@aparcar Any update regarding this PR?

rany2:
Could you rerun the test now? I believe I've fixed it now, wasn't aware that I needed to add a config item in the configtest.py

asu/openapi.yml
```diff
@@ -265,6 +265,13 @@ components:
             Ability to specify filesystem running on device. Attaching this
             optional parameter will limit the ImageBuilder to only build
             firmware with that filesystem.
+        rootfssize:
```
aparcar, line 268: Please specify a maximum, see the specific parameter names [here](https://swagger.io/docs/specification/data-models/data-types/#numbers)

misc/config.py
```diff
@@ -16,4 +16,7 @@
 # where to store JSON files
 JSON_PATH = Path.cwd() / "public/json/v1/"
 
+# maximum custom rootfs size in MB
+MAX_CUSTOM_ROOTFS_SIZE_MB = 14748
```
aparcar, line 20: This isn't required when adding it to the openapi spec

tests/conftest.py
```diff
@@ -84,6 +84,7 @@ def app(test_path, redis_server):
             "CACHE_PATH": test_path,
             "TESTING": True,
             "UPSTREAM_URL": "http://localhost:8001",
+            "MAX_CUSTOM_ROOTFS_SIZE_MB": 14748,
```
aparcar, line 87: Neither is this required.

asu/build.py
```diff
@@ -300,6 +300,42 @@ def download_file(filename: str, dest: str = None):
             cache_workdir / ".config",
         )
 
+    rootfssize = None
+    if req.get("rootfssize") and isinstance(req.get("rootfssize"), int):
+        if 1 < req.get("rootfssize") <= req["max_custom_rootfs_size_mb"]:
+            rootfssize = req.get("rootfssize")
+        elif req.get("rootfssize") > req["max_custom_rootfs_size_mb"]:
+            rootfssize = req["max_custom_rootfs_size_mb"]
```
aparcar, line 308: Use openapi max and minimum to verify the data instead of checking it here afterwards.

aparcar (changes requested):

asu/openapi.yml
rany2: Shouldn't the maximum be configurable by the server operator? I mainly thought to add it to prevent an asu server from building an image that would take up too many resources

rany2: @aparcar I've done so in the latest commit however I'm still not sure how to have it configurable via openapi.yml

rany2:
Currently max is set in openapi.yml, if there is a better solution than this I'm open to it but I guess this is fine for now.

aparcar:
Yea sorry I think you're right and it should be part of the config file. Mind switching it back? I'll then merge it

rany2:
@aparcar After reading up on connexion I realized that it is in fact possible for us to specify max from config. You could check how I did this in https://github.com/openwrt/asu/pull/416/commits/da91d234b56dbae2d5ae0b6796e8f13f7c2d6f4d, if you prefer how I did it previously, I am willing to bring it back!

misc/config.py
```diff
@@ -16,4 +16,7 @@
 # where to store JSON files
 JSON_PATH = Path.cwd() / "public/json/v1/"
 
+# maximum custom rootfs size
+MAX_CUSTOM_ROOTFS_SIZE_MB = 131072
```
aparcar, line 20: 131GB of rootfs? Could you please check the size of the resulting compressed images and how long it takes to build those? It's an obvious DOS attack to create some large images and thereby filling op the server storage.

@dangowrt can you think of a better way to handle this? Usually I'd expect to have a limit of 1GB to install binaries and then another data partition.

aparcar: I'd wait for a comment of @dangowrt since he's done some prior work on partition managing. I'm fine with a max of say 512MB or 1024MB but more than that seems excessive for installed binaries.

rany2: @dangowrt Any updates?

mjnck:
Any updates? Maybe just set CONFIG_TARGET_ROOTFS_PARTSIZE to size of rootfs on openwrt asu is running on?

rany2:
@mjnck This is what this would allow. The parameter is necessary so that acu could request the size that it is running on.

aparcar:
I can this with three different sizes and the time it takes to build those images is terrible. We need to speed this up over at upstream before considering it. Call me boring but with something like this in place it's trivial to DOS the server.

```
104: real	0m35.339s
512 :real	1m8.780s
1024: real	1m50.652s
```

rany2:
@aparcar IIRC the primary reason why this is so slow is because of squashfs, maybe we should only allow this feature if squashfs was requested?

mjnck:
Maybe just add summary of all included packages size to default 104mb size. At least it should prevent build errors due to lack of free space

joshenders:
Happy to see progress here as this has been a thorn in my side for years but has there been any discussion on whether it would make sense to package the x86 release as a small initrd with an installer?

joshenders:
@aparcar Do you have any suggestions on what the best path forward here might be for @rany2?

aparcar:
I just merged [this](https://github.com/openwrt/openwrt/commit/7b7edd25a571568438c886529d3443054e02f55f) which allows to specify ROOTFS_PARTSIZE when using the ImageBuilder. Do you mind reworking your PR to make use of that, please?

I'm trying to get away from any custom .config manipulation as I'm planing to run all ImageBuilders inside a Bodman container to improve security. Right now malicious packages can do harm, containerized they are much more limited.

rany2:
@aparcar definitely, this change will make things much easier for me. I was never a fan of .config manipulation in the first place.

ascendbeing:
> I just merged [this](https://github.com/openwrt/openwrt/commit/7b7edd25a571568438c886529d3443054e02f55f) which allows to specify ROOTFS_PARTSIZE when using the ImageBuilder.
Is this in any way responsible for why ASU/FW Selector/attended sysup doesn't work now?
It'll first claim that various packages are unsupported. For me this was  htop, dnsmasq-full, kmod, libkmod, htop, miniupnpd-nftables, ethtool-full, etc.
If you clear that hurdle by getting all the packages it names as unsupported, next thing that happens is it says impossible package selection and feeds you a log.
In that log what I was seeing is nftables or nftables-json isn't being installed which I think means that the package itself isn't found on the package downloads anymore. Because all those packages I named off, I think all or most of them have no description upon update of the opkg list. They don't show up on available packages, only installed packages. Even nftables isn't on available.
I'm seeing this same thing on both ath79 and kirkwood targets. To resummarize it seems like packages are missing on the package download pages. Nftables or nftables-json being the big one. Possibly also or just the packages I rattled off about above.
Thanks @aparcar I will edit this post to let you know what revision I'm on it's been broken for 1-2 days I think but I'll show you the revision number in an edit

ascendbeing:
@aparcar I further looked into it and https://downloads.openwrt.org/snapshots/packages/mips_24kc/ there's no dnsmasq-full or nftables or nftables-json in base or routing or packages. There's currently a build job going for this package target, maybe for mips_24kc this will post the missing packages? 
I haven't manually checked arm_xscale but that's the other one where this behavior was happening so I suspect it's also missing those packages and notables/-json
I've never seen this happen, where a non kmod package wouldnt be built or would be seemingly removed, when it isn't dependent on a particular kernel version.

aparcar:
@ascendbeing There are [upstream issues](https://downloads.openwrt.org/snapshots/faillogs/mipsel_24kc/base/dnsmasq/full/compile.txt) with dnsmasq-full. Please create a dedicated issue over at openwrt/openwrt or see if someone else reported the bug already. Snapshots are not recommended for anyone who is not actively developing or following the source changes, please switch to a stable release.

ascendbeing:
> 

@aparcar Thanks for the reply I have confirmed that the latest run of master mips_24kc packages build didn't upload the missing packages and in fact if you read the "compile" log of that build, it shows numerous packages failing, including nftables and nftables-json. Apparently what's happening there is instead of failing when major packages fail, it's proceeding as if despite major, default packages failing, it's good to continue the build and overwrite the available packages without those packages.

Re: Dnsmasq. I'm not a stickler for it. I was just telling you a bunch of packages are failing during build then erased because the build job continues like it's cool even for default included packages for the snapshot images.

I'm technically a developer. I have contributed a fix for Kirkwood not retaining config files when you do a sysupgrade. https://github.com/openwrt/openwrt/pull/11692 it's a 2 line fix each for ipq40xx and Kirkwood in their respective linksys.sh files which hasn't been implemented. I did test Kirkwood against the latest change to that file btw which I didn't post there, works fine with my fix 

Should I file a bug report on the packages GitHub? Now that it's clear that the problem is a bunch of packages failing to build and then it proceeding as if it's good and replacing working packages with a void of them because the failure didn't drop the build in it's tracks. I wasn't trying to single you out in particular, i started yesterday with the assumption ASU changed, but it's a building packages issue.

asu/build.py
```diff
@@ -308,6 +308,9 @@ def download_file(filename: str, dest: str = None):
         f"EXTRA_IMAGE_NAME={packages_hash}",
         f"BIN_DIR={req['store_path'] / bin_dir}",
     ]
+    rootfs_size_mb = req.get("rootfs_size_mb")
+    if rootfs_size_mb:
```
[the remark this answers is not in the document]
aparcar: This runs on 3.8 anyway

rany2:
@aparcar I'm not sure why the tests are failing. Could you provide some guidance on this? I'm lost.

tests/conftest.py
```diff
@@ -84,6 +84,7 @@ def app(test_path, redis_server):
             "CACHE_PATH": test_path,
             "TESTING": True,
             "UPSTREAM_URL": "http://localhost:8001",
+            "MAX_CUSTOM_ROOTFS_SIZE_MB": 100,
```
rany2, line 87: Would it be fine if this was reverted then?

aparcar: Yea not needed

aparcar:
To github.com:openwrt/asu.git
   98c58b8..e87941f  main -> main

Merged, thanks you! Sorry for the long wait

PR #428, opened 2022-12-29
fixed release_date in janitor.py

### LINKED ISSUE #427: ASU own Server (Docker): Build working with Swagger but not with AUC
Followed the Readme in order to run my own Sysupgrade server in Docker. Testing with Swagger (i.e. the built-in API test, in my case `http://10.39.39.246/ui/#/default/asu.api.api_v1_build_post` with the following test data 

```
{
  "packages": [
    "auc",
    "luci-app-attendedsysupgrade"
  ],
  "profile": "8dev_carambola2",
  "target": "ath79/generic",
  "version": "22.03.2"
}

```
works well - the image gets built and I can see all the files in the ...public/store/hash/ subdirectory.

However, trying **auc** from an OpenWrt Router, auc throws a blunt `No data available (61)` - I've set the Server URL to 

```
uci set attendedsysupgrade.server.url='http://10.39.39.246'
uci commit
auc
```

(the Docker stack is answering on port 80). Looking at the screen output of  the docker stack, nothing happens - it looks like auc is not even requesting anything. I can however do a wget/curl from the router to the Docker Server (e.g. `wget http://http://10.39.39.246/` will grab the index.html)

Is there anything else that I need to tell auc in order to make it happy ?

uci shows
```
attendedsysupgrade.server=server
attendedsysupgrade.server.url='http://10.39.39.246'
attendedsysupgrade.client=client
attendedsysupgrade.client.upgrade_packages='1'
attendedsysupgrade.client.auto_search='0'
attendedsysupgrade.client.advanced_mode='1'
```
auc works well with `https://sysupgrade.openwrt.org`

Any help is appreciated - Many thanks in advance!

onemarcfifty:
The janitor would fail if a branch node contained a date value. As a result, the branches.json, latest.json and overview.json files are never created and therefore acu and/or luci-app-attendedsysupgrade would not be able to update from the asu instance.

Converting json node values to string by default fixes the issue.

This fixes #427 

Signed-off-by: Marc Ahlgrim <marc@onemarcfifty.com>

### DIFF
--- a/asu/janitor.py
+++ b/asu/janitor.py
@@ -419,11 +419,18 @@ def update_meta_json():
         json.dumps(
             current_app.config["OVERVIEW"],
             indent=2,
+            sort_keys=False,
+            default=str
         )
     )
 
     (current_app.config["JSON_PATH"] / "branches.json").write_text(
-        json.dumps(list(branches.values()))
+        json.dumps(
+            list(branches.values()),
+            indent=2,
+            sort_keys=False,
+            default=str
+        )
     )
 
     (current_app.config["JSON_PATH"] / "latest.json").write_text(


onemarcfifty:
Hi @aparcar @dangowrt - would you please mind approving the CI workflow ? I am a first time contributor in this repo. Many thanks !

asu/janitor.py
```diff
@@ -419,11 +419,18 @@ def update_meta_json():
         json.dumps(
             current_app.config["OVERVIEW"],
             indent=2,
+            sort_keys=False,
```
aparcar, line 422: Why not sort them? Isn't that better for reproducibility?

aparcar (changes requested):

aparcar:
Thanks for your contribution and the quick fix!

asu/janitor.py
onemarcfifty: I thought it might make more sense to have them in the same order like the branches.yml (if you have one) and also in the same order like the structure that is defined in the source code (`current_app.config["OVERVIEW"]`). But yes - we could sort them

aparcar: Good point. Arrays should be sorted at any point but dicts are.

aparcar:
Could you please add the PR text to the commit message?

aparcar (approved):

onemarcfifty:
added PR text to commit

aparcar:
Awesome thank you very much!

PR #460, opened 2023-02-18
api: allow to prefix packages with a +

aparcar:
When adding a + to the beginning of a package name it will modify the order of packages and thereby fix dependency issues of OPKG.

This is more of a advanced user feature.


aparcar:
@dangowrt thoughts?

dangowrt:
Sounds like a viable workaround to the problems with opkg. Of course, it would be better to not even need this kind of hacks in first place, but I also don't see anything problematic about this specific solution.

PR #478, opened 2023-04-07
build: use podman for image creation

aparcar:
- [x] custom package repositories
- [x] update documentation
- [x] update docker-compose or replace with Podman-compose
- [x] add cleanup mechanism
- [ ]


asu/build.py
```diff
@@ -29,187 +22,90 @@ def build(req: dict):
...
-            req["upstream_url"]
-            + "/"
-            + req["branch_data"]["path"].format(version=req["version"])
-            + "/targets/"
-            + req["target"]
-            + "/"
-            + filename
-        )
+    podman = PodmanClient(
+        base_url="ssh://user@localhost:50622/tmp/podman.sock",
```
dangowrt, line 67: SSH to localhost looks strange to me. Isn't there a URI format which would just use the local socket (ie. something like `file:///tmp/podman.sock`)?

aparcar: This is a bit of an awkward thing when running Bodman on MacOS, it's running in its own VM to have a Linux Kernel running. The official Podman is incredibly slow and using alternatives are focused on Docker. Long story short I'll make this user configurable.

PR #488, opened 2023-04-20
Fix typo where there should be a set and not a list in request packages

aparcar:


aparcar:
@1715173329 could you please have a look?

1715173329:
I can confirm it fixed my issue, thank you!

Issue #628, opened 2023-10-23
kmod-switch-rtl8366s and similar are needed for various ath79 devices when upgrading to 23.05.0


d1b:
Hi, I used `auc`  to upgrade an wndr3800 from 22.03.5 to 23.05.0 & as `kmod-switch-rtl8366s` is now needed to configure the switch, the device wasn't working as it was supposed to.  This seems to relate to https://github.com/openwrt/openwrt/commit/6e0f0eae5b48bddead035d9dd89b710c10c50337  & https://github.com/openwrt/openwrt/commit/f3bb1eea32c728470563e7e0df8bd905a913e074 .

Please consider adding a workaround so that `kmod-switch-rtl8366s`  or `kmod-switch-rtl8366rb` is pulled in as appropriate.

d1b:
I'll have a look at raising a pr tomorrow :-) .

PR #635, opened 2023-10-26
Workaround kmod-switch-rtl8366s & kmod-switch-rtl8366rb being needed for specific devices

d1b:
Fix: #628


aparcar:
Thanks I added your commit and then refactored it to a separate file!

PR #657, opened 2023-11-17
branches: add 23.05.2

dgilman:
I did not bump the date - is that expected? The documentation is not clear what it is used for, or supposed to be used for.


beolis:
Thanks @dgilman & @aparcar! It's upgrade o'clock here.

PR #727, opened 2024-02-03
config: allow to modify operations timeout

a-gave:
This introduces an additional option to modify the operations timeout (currently defaulted to '10m')

The reason is to allow an installation of asu even on a machine that has low bandwidth availability, or a fair amount of latency.
Under these conditions asu may work fine, however the installation process may timeout during the profiles.json update process (unless the amount of available branches/releases is reduced)
In addition, even a single build may timeout if more than 10 minutes are required between downloading the imagebuilder corresponding to a request and compiling it.

### DIFF
--- a/asu/janitor.py
+++ b/asu/janitor.py
@@ -9,6 +9,16 @@ from rq import Queue
 from asu import __version__
 from asu.common import get_redis_client, is_modified
 
+from pathlib import Path
+import importlib.util
+
+for config_file in ["/etc/asu/config.py", "misc/config.py"]:
+    if Path(config_file).exists():
+        spec = importlib.util.spec_from_file_location("spec", config_file)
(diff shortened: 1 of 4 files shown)


asu/janitor.py
```diff
@@ -9,6 +9,16 @@
 from asu import __version__
 from asu.common import get_redis_client, is_modified
 
+import os.path
+import importlib.util
+
+for config_file in ["/etc/asu/config.py", "misc/config.py"]:
+    if os.path.exists(config_file):
```
aparcar, line 16: Please use pathlib instead of os.path

Neustradamus:
What is the status of this PR?

a-gave:
This referred to an older version of the code, and now seems unnecessary. I'm closing it.

PR #728, opened 2024-02-03
build: allow to define additional configs

a-gave:
This introduces the possibility of passing additional configs to the request.

Only the few not related to kernel/packages, as they are not recompiled by the imagebuilder, would make sense. 
For example, to request a specific name/number for a firmware image: 
```
CONFIG_VERSION_DIST=MyRouterOS
CONFIG_VERSION_NUMBER=1.6
```

Or to request a format other than the default format for the output firmware image, for example:
```
# CONFIG_TARGET_ROOTFS_SQUASHFS is not set
CONFIG_TARGET_ROOTFS_JFFS2=y
```

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -150,6 +150,23 @@ def build(build_request: BuildRequest, job=None):
             },
         )
 
+    if build_request.configs:
+        log.debug("Found extra configs")
+        configs = ""
+        for config in build_request.configs:
+            configs += f"{config}\n"
+
+        (bin_dir / ".config_local").write_text(configs)
+
+        mounts.append(
+            {
+                "type": "bind",
+                "source": str(bin_dir / ".config_local"),
+                "target": "/builder/.config_local",
+                "read_only": True,
+            }
+        )
+
     log.debug("Mounts: %s", mounts)
 
     container = podman.containers.create(
@@ -242,6 +259,21 @@ def build(build_request: BuildRequest, job=None):
 
     packages_hash: str = get_packages_hash(manifest.keys())
     log.debug(f"Packages Hash: {packages_hash}")
+
+    if build_request.configs:
+        log.info("Appling local configs")
+        returncode, job.meta["stdout"], job.meta["stderr"] = run_cmd(
+            container,
+            [
+                "sh",
+                "-c",
+                (
+                    "for config in $(grep '#' .config_local | awk '{print $2}'); do sed -i 's/'$config'.*//' .config ; done; cat .config_local >> .config"
+                ),
+            ],
+        )
+        if returncode:
+            report_error(job, "Could not apply local configs")
 
     job.meta["build_cmd"] = [
         "make",
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -17,6 +17,10 @@ from asu.util import (
     get_request_hash,
 )
 
+import re
+
+configs_extract = r"^# (CONFIG_[\w_.-]) is not set|(CONFIG_[\w_.-]+)"
+
 router = APIRouter()
 
 
@@ -70,6 +74,12 @@ def validate_request(build_request: BuildRequest) -> tuple[dict, int]:
 
     if build_request.distro not in get_distros():
         return validation_failure(f"Unsupported distro: {build_request.distro}")
+
+    if build_request.configs:
+        for config in build_request.configs:
+            config_key = re.search(configs_extract, config)
+            if config_key[0] not in settings.configs_allowed:
+                return validation_failure("Illegal config requested: {config_key[0]}")
 
     branch = get_branch(build_request.version)["name"]
 
--- a/asu/util.py
+++ b/asu/util.py
@@ -97,6 +97,18 @@ def get_file_hash(path: str) -> str:
     return h.hexdigest()
 
 
+def get_configs_hash(configs: list) -> str:
+    """Return sha256sum of configs list
+    Duplicate configs are automatically removed and the list is sorted to be
+    reproducible
+    Args:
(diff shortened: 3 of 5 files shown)


aparcar:
Please fix the failing test.

Neustradamus:
What is the status of this PR?

aparcar:
I'm not a huge fan to allow arbitrary changes but since we moved into containers, it might be fine. Is anyone interested in using this?

a-gave:
Hi, rebased on current Git HEAD, let me know if there is any interest on this, , and/or if there are things that should be improved. I would still be interested.

aparcar:
Should we add a configuration which sets specific CONFIG options which are allowed? I'm a bit unsure this doesn't cause any issues down the road. For example you could simply bypass the root filesystem size limitation of rootfs_size_mb and set it to 100GB, filling up the servers storage...

a-gave:
> Should we add a configuration which sets specific CONFIG options which are allowed? I'm a bit unsure this doesn't cause any issues down the road. For example you could simply bypass the root filesystem size limitation of rootfs_size_mb and set it to 100GB, filling up the servers storage...

Thanks, this make a lot of sense, I've pushed some modifications to have an allow-list of configs, stored in settings.

a-gave:
pushed fixes on formatting (https://github.com/openwrt/asu/actions/runs/12634720904)

aparcar:
@efahl what do you think about this?

efahl:
I'd say before proceeding that a clear case needs to be made for the specific config vars of interest, and that each would need a good description as to effects and possible issues.  (If, for example, CONFIG_TARGET_ROOTFS_TARGZ is disabled, will that create huge uncompressed images that fill the server?)  And, if there is a good case to be made for specific ones, they should be incorporated into the imagebuilder first, exposing them just as ROOTFS_PARTSIZE and EXTRA_IMAGE_NAME are...

But, I'm curious what ASU clients would use this?

And for some of the values, I think it would cause the client some problems.  For example, if you change the rootfs to `jffs2` via a CONFIG var, then all of `auc`, `owut` and LuCI app will not know what to do because they'll still be looking in profiles.json for a `squashfs` one.  Also with rootfs type, there's the issue that the imagebuilders must have been created with the proper rootfs variants in the first place (see how `owut` handles this with its `--fstype` option validation).

asu/build.py
```diff
@@ -243,6 +260,21 @@ def build(build_request: BuildRequest, job=None):
...
 
+    if build_request.configs:
+        log.info("Appling local configs")
+        returncode, job.meta["stdout"], job.meta["stderr"] = run_cmd(
+            container,
+            [
+                "sh",
+                "-c",
+                (
+                    "for config in $(grep '#' .config_local | awk '{print $2}'); do sed -i 's/'$config'.*//' .config ; done; cat .config_local >> .config"
```
efahl, line 271: Why not do this filtering when you are creating the .config_local file?

asu/routers/api.py
```diff
@@ -17,6 +17,10 @@
     get_request_hash,
 )
 
+import re
+
+configs_extract = r"^# (CONFIG_[\w_.-]) is not set|(CONFIG_[\w_.-]+)"
```
efahl, line 22: Shouldn't the second clause also be anchored?  As is, the second clause matches anywhere in the line.

asu/routers/api.py
```diff
@@ -71,6 +75,12 @@ def validate_request(build_request: BuildRequest) -> tuple[dict, int]:
     if build_request.distro not in get_distros():
         return validation_failure(f"Unsupported distro: {build_request.distro}")
 
+    if build_request.configs:
+        for config in build_request.configs:
+            config_key = re.search(configs_extract, config)
+            if config_key[0] not in settings.configs_allowed:
+                return validation_failure("Illegal config requested: {config_key[0]}")
```
efahl, line 82: Should be an f-string to evaluate the expression.

asu/util.py
```diff
@@ -97,6 +97,18 @@ def get_file_hash(path: str) -> str:
...
+def get_configs_hash(configs: list) -> str:
+    """Return sha256sum of configs list
+    Duplicate configs are automatically removed and the list is sorted to be
+    reproducible
+    Args:
+        configs (list): list of configs
+    Returns:
+        str: hash of `req`
+    """
+    return get_str_hash(" ".join(sorted(list(set(configs)))))
```
efahl, line 109: You could leave out that extra `list` call, just say `sorted(set(` as that has the same effect and eliminates an extra list-ification.

a-gave:
> I'd say before proceeding that a clear case needs to be made for the specific config vars of interest, and that each would need a good description as to effects and possible issues. (If, for example, CONFIG_TARGET_ROOTFS_TARGZ is disabled, will that create huge uncompressed images that fill the server?) And, if there is a good case to be made for specific ones, they should be incorporated into the imagebuilder first, exposing them just as ROOTFS_PARTSIZE and EXTRA_IMAGE_NAME are...

Thanks also for the reviews, I agree with that, and with considerations on asu. I will continue for a while to use a fork that does that. But I'm actually interested in the first two options, the others were added as possibilities that were present, but didn't generate interest, so I think they are not used, and at the limit as you say should be added to the imagebuilder. I had missed then that the use of JFF2 is also discouraged. [0]

[0] https://openwrt.org/docs/techref/filesystems#can_we_switch_the_filesystem_to_be_entirely_jffs2

efahl:
If you do end up submitting something for imagebuilder to directly support your usage, tag me on the PR.  I'd be glad to review and if it's accepted, then get it into ASU.

Issue #826, opened 2024-05-11
Max ROOTFS_PARTSIZE clamped to unrealistic value


efahl:
Doing some testing, requested an image from the ASU build server with a 256 MB root partition using this request
```
{
    "client": "owut/24.0.0",
    "target": "x86/64",
    "profile": "generic",
    "version": "SNAPSHOT",
    "filesystem": "ext4",
    "rootfs_size_mb": 256
}
```
 and it fails with 
```
Status: 400
Detail: 256 is greater than the maximum of 100 - 'rootfs_size_mb'
```
I build images all the time with 512MB root partitions, and the default for all builds in config/Config-images.in is 104 with two exceptions: target mediatek (448) and loongarch64 (232).  Having it limited to 100 seems incorrect, maybe 1000/1024 would be a better ceiling?

I think the culprit is here:
https://github.com/openwrt/asu/blob/101831b3a04d7af780f0540f52082873dd1933d8/asu/asu.py#L38
but could be here:
https://github.com/openwrt/asu/blob/101831b3a04d7af780f0540f52082873dd1933d8/misc/config.py#L17
(or both?)

PR #827, opened 2024-05-12
build: increase MAX_CUSTOM_ROOTFS_SIZE_MB to a reasonable value

efahl:
Current value of 100 is less than default when building OpenWrt (104). Increase it to 1024 to allow large root file systems on devices that can accommodate them.

https://github.com/openwrt/openwrt/blob/61e8728d86d7c11e1a5adab4bd37dae6b3b6cf2b/config/Config-images.in#L306

Closes: https://github.com/openwrt/asu/issues/826

### DIFF
--- a/asu/asu.py
+++ b/asu/asu.py
@@ -35,7 +35,7 @@ def create_app(test_config: dict = None) -> Flask:
         ALLOW_DEFAULTS=bool(getenv("ALLOW_DEFAULTS", False)),
         ASYNC_QUEUE=True,
         BRANCHES_FILE=getenv("BRANCHES_FILE"),
-        MAX_CUSTOM_ROOTFS_SIZE_MB=100,
+        MAX_CUSTOM_ROOTFS_SIZE_MB=1024,
         REPOSITORY_ALLOW_LIST=[],
         BASE_CONTAINER="ghcr.io/openwrt/imagebuilder",
     )
--- a/misc/config.py
+++ b/misc/config.py
@@ -14,7 +14,7 @@ STORE_PATH = Path.cwd() / "public/store/"
 JSON_PATH = Path.cwd() / "public/json/v1/"
 
 # maximum custom rootfs size
-MAX_CUSTOM_ROOTFS_SIZE_MB = 100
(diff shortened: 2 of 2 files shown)


aparcar:
Did you test how long it takes to have a rootfs of one gigabyte?

efahl:
> Did you test how long it takes to have a rootfs of one gigabyte?

Here's the results from doing image builder runs of x86/64 generic at three root fs sizes:

`$ time make image PROFILE=generic ROOTFS_PARTSIZE=512`

I also checked the resulting size of `openwrt-x86-64-generic-ext4-combined-efi.img.gz` for each build, as that seems to be the biggest of the images produced. Times are the worst case of three runs at each size, rounded up to nearest second (variation was always < 1s, so one run would have sufficed).

| ROOTFS_PARTSIZE= | real | user | img size |
| ---------------: | ----: | ----: | --------: |
|  104 |  26s |  18s |  12M |
|  512 |  48s |  25s |  13M |
| 1024 |  74s | 33s |  13M |

(Those `dd`s take forever, don't they?)

On the production ASU server, I measure the "104" builds to take about 90s, so my build platform is running about 3x faster than the server.  Scaling that 74s * 3 = 225s = 3.75m, call it ~4 minutes...

aparcar:
@dangowrt what do you think about this?

efahl:
Just did a build using 256 MB, seems to work just fine.  My time estimate was a bit low, 3.5 minutes for this one, but otherwise 👍🏼
```
Image: https://sysupgrade.openwrt.org/store/1c2a8746a0b600c998b9bb7b873ef34c/openwrt-f2f3d6a0adb2-x86-64-generic-squashfs-combined-efi.img.gz
Downloaded https://sysupgrade.openwrt.org/api/v1/build/1c2a8746a0b600c998b9bb7b873ef34c to /tmp/owut-build-status.json
Status: 200
Detail: done
Build completed in 210 seconds.
build_at       = 2024-05-12T11:49:39.000000Z
version_number = SNAPSHOT
version_code   = r26300-da0cd9d764
rootfs_size_mb = 256
```

PR #868, opened 2024-06-26
config: add entry to package_changes

efahl:
'auc' has been removed and replaced by 'owut', reflect that in the 'package_changes'

https://github.com/openwrt/packages/commit/3a998e10218c318511b41739f276e572c1ede967

### DIFF
--- a/asu/branches.yml
+++ b/asu/branches.yml
@@ -90,6 +90,9 @@ branches:
     versions:
       - SNAPSHOT
     package_changes:
+      - source: auc
+        target: owut
+        revision: 26792
       - source: libustream-wolfssl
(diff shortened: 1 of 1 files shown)


aparcar:
@dangowrt your blessings?

dangowrt (approved):
I hope it will be obvious to users that `owut` is the new `auc`. Maybe we should install a symlink called `auc` pointing to `owut`...?

efahl:
> Maybe we should install a symlink called `auc` pointing to `owut`...?

Or maybe a new `auc` 1-line script?
 echo "auc removed, use 'owut --help' to see new command line options"

PR #905, opened 2024-07-22
routers/api: add constraints to API fields

efahl:
Use pydantic's Field to add constraints to 'defaults' (string length <= 20480) and 'rootfs_size_mb' (int 1 <= value <= 1024) 'package_changes'

Reference:
https://docs.pydantic.dev/latest/concepts/fields/#numeric-constraints

Signed-off-by: Eric Fahlgren @efahl

### DIFF
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -1,9 +1,9 @@
 import logging
-from typing import Annotated, Union
+from typing import Annotated, Optional
 
 from fastapi import APIRouter, Header
 from fastapi.responses import RedirectResponse, Response
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from asu.build import build
 from asu.config import settings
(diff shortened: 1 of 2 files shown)


asu/routers/api.py
```diff
@@ -20,9 +20,21 @@ class BuildRequest(BaseModel):
     packages: Union[list, None] = []
     profile: str
     packages_versions: dict = {}
-    defaults: Union[str, None] = None
+    defaults: Optional[
+        Annotated[
+            str,
+            Field(default=None, max_length=20480, description="asu-defaults script"),
```
aparcar, line 26: It's called uci-defaults, just to avoid confusion :)

asu/routers/api.py
```diff
@@ -20,9 +20,21 @@ class BuildRequest(BaseModel):
...
+        ]
+    ] = None
     client: Union[str, None] = "unknown/0"
-    rootfs_size_mb: Union[int, None] = None
+    rootfs_size_mb: Optional[
+        Annotated[
+            int,
+            Field(
+                default=None, gt=0, le=1024, description="Resize the root file system."
```
aparcar, line 34: Could you add this to the config.py please to make it configurable?

asu/routers/api.py
aparcar, line 26: Could you add this to the config.py please to make it configurable?

aparcar:
Thank for this PR, looks great!

efahl:
While I'm poking around here...

What do you think about changing the other `Union[type, None]` declarations into `Optional[type]`? (At least for the ones right around the ones I'm changing.)  As I recall, `Optional` was added to typing way back in 3.8, and it seems to be the more common idiom these days.

asu/routers/api.py
efahl: I will go back to old openapi.yml and steal the comments from there.  Not sure why I didn't think of that sooner...

aparcar:
Please go ahead, this typing stuff is rather new for me

asu/routers/api.py
efahl: Put the yaml comments into the new code.

asu/routers/api.py
efahl: All constants removed and now come from `settings...`, including the description for `defaults`.

aparcar:
Excellent thank you very much. Will deploy this version tomorrow

PR #909, opened 2024-07-23
api: restore 'version_code' for auc and owut

efahl:
Restore the 'version_code' field as it's used by both auc and owut to verify the resulting build.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -65,11 +65,11 @@ def build(req: dict, job=None):
 
     version_code = re.search('Current Revision: "(r.+)"', job.meta["stdout"]).group(1)
 
-    if "version_code" in req:
-        if version_code != req.get("version_code"):
+    if requested := req.get("version_code"):
+        if version_code != requested:
             report_error(
                 job,
-                f"Received incorrect version {version_code} (requested {req['version_code']})",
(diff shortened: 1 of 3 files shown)


asu/util.py
```diff
@@ -115,7 +115,7 @@ def get_request_hash(req: dict) -> str:
             [
                 req.get("distro", ""),
                 req.get("version", ""),
-                req.get("version_code", ""),
+                req.get("version_code", "") or "",
```
aparcar, line 118: I think this isn't needed since .get defaults to "" already

aparcar (changes requested):

asu/util.py
efahl: It is needed currently, because if you do
```
req["version_code"] = None       # The current default.
x = req.get("version_code", "")  # x is now None and 'join' crashes.
```
I'll rework, and see what happens if we just leave the default value out...

efahl: Ok, seems I keep confusing typing's notion of optional and pydantic's optional...

```
field: str  = ""     # Pydantic optional because it has a default value
field: Optional[str] # Pydantic required because no default value
```

aparcar (approved):

PR #910, opened 2024-07-25
build: retain user input order in package list

efahl:
This allows users to work around an opkg bug that causes package installation failure when the package list is sorted.  For one specific example, see https://forum.openwrt.org/t/owut-openwrt-upgrade-tool/200035/61.

- Firmware selector needs to have its sort removed.
  https://github.com/openwrt/firmware-selector-openwrt-org/blob/master/www/index.js#L618
- I already updated owut. https://github.com/efahl/owut/commit/d8af324ccad0c93f0d6a9fd1e72236e55d480a7c
- Haven't looked at LuCI ASU yet.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -91,7 +91,7 @@ def build(req: dict, job=None):
 
     if req.get("diff_packages"):
         req["build_cmd_packages"] = diff_packages(
-            set(req["packages"]), default_packages | profile_packages
+            req["packages"], default_packages | profile_packages
         )
         log.debug(f"Diffed packages: {req['build_cmd_packages']}")
     else:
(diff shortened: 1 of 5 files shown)


aparcar:
This really lowers the chances of an image being cached so this must be removed once APK is used as a replacement...

efahl:
> This really lowers the chances of an image being cached

I think it causes the opposite problem, actually...  

Since the hash keys are still using the sorted lists, the input package order doesn't change the hash keys (I checked, I should add some tests).  But, if someone gets an error on firmware selector due to this problem, then swaps package order, the hash key is the same and they get back the cached-broken build result...  We shouldn't see this situation with LuCI, auc or owut, as the user can't change the order (at least not easily or obviously).

PR #911, opened 2024-07-25
metrics: use Redis timeseries

aparcar:
Just for testing, but this could make things simpler...


aparcar:
@efahl what do you think about adding stats via time series? or do you have a different idea? I'm not super happy with using Prometheus since the "complexity" of the metrics used up more performance than the building itself at some point. 

Maybe we could also create a JSON file once per out and render that manually? I don't know...

efahl:
Ok, so what are our goals with the metrics?  I recall some university researcher looking at the numbers a few months back (I found it https://forum.openwrt.org/t/is-this-download-numbers/189455), but that's sort of a side quest.  (Or maybe it is useful, if it can be used by someone to get funding.)

As far as I'm concerned, I'd like to see internal builder cache hits vs misses more than anything (to better help us diagnose issues).  Binning by device/version/platform/whatever is probably useful to devs for setting global direction, but that is a report you could generate on-demand (say for the "monthly" dev meetings).

In any case, I would think weekly updates would be more than enough, any realtime stats are sort of a waste of CPU.

PR #912, opened 2024-07-25
test_api: verify request hash with different package ordering

efahl:
Add test cases with both diff_packages true and false, with the input package list both unordered and sorted.  Request hash must be identical regardless of package list order.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -104,6 +104,16 @@ def test_api_build_version_code_bad(client):
     )
 
 
+base_packages_diff = (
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -121,17 +131,68 @@ def test_api_build_diff_packages(client):
     data = response.json()
 
     # TODO shorten for testing
```
aparcar, line 133: this could now be removed :)

aparcar:
@efahl please fix the one failing checksum (8731372f84b0022c070e6127bad24eb2)

PR #969, opened 2024-09-10
api: add preliminary builder statistics

efahl:
Give clients the ability to check how busy the server is, so they can give diagnostics and/or refuse to queue up a job if there are too many build requests in the queue.

@aparcar , what do you think of this one?  I want to be able to check if there are too many jobs already in process and then say "sorry, you'll be waiting three hours for this build, not gonna do it..."


aparcar:
Issue here was Podman to "die" on the server and no longer do anything meaningful, i.e. not starting new containers which actually work on jobs. Normally I've never seen queues longer than 10. This fix (even though best intentions) don't fix the underlying issue...

efahl:
Right, my logic is:
1) Client (owut, FS) requests queue length before submitting build.
2) If not 200, then something is wrong and supply some diagnostics: "ASU server down blah blah".
3) If 200, then 
4) If queue_length < say 10, then just go ahead.
5) If queue_length too big (it was 104 this morning), then tell user "it will be 3*queue_length minutes before your job starts" and abort submission of job.

My hope is to warn users that something is wrong, so they don't just start posting new issues when it's obvious that there is something going on...  It doesn't fix root causes, but it lets users know they shouldn't expect things to be working.

efahl:
Oh, and also would be nice for FS/owut/LuCI to say "The server is currently building N jobs..."  Instantaneous status when you want to request a build.

asu/routers/api.py
```diff
@@ -271,3 +271,15 @@ def api_v1_build_post(
     response.status_code = status
 
     return content
+
+
+@router.get("/builder_stats")
+def api_v1_builder_stats():
```
aparcar, line 277: Let's call this just api/v1/stats and I'll merge it?

efahl: Will do.

aparcar:
Can't merge this as is, please rebase

PR #979, opened 2024-09-19
tests: fix deprecation warnings

efahl:
Use the new pydantic 'model_copy' method on BaseModel to eliminate deprecation warnings when running the tests.

(Picking away at some little stuff before I start submitting some bigger ones...)


aparcar:
I love that, thanks!

PR #980, opened 2024-09-19
asu/update: avoid crashes on non-existent or obsolete versions

efahl:
Add some logic to gracefully ignore updates on old versions that lack the proper data to be supported.

Every time I did a cold start, I'd get a bunch of errors on the worker:
```
  File "/app/asu/update.py", line 26, in update_targets
    version_path = branch_data["path"].format(version=version)
                               ^^^^^^
KeyError: branch_data has no key "path"
```
Stuck an assertion in and turns out to be the `19.07` stuff,
```
AssertionError: No path for 19.07.7
```

Not sure if it happens in production, but this fixes it and cleans up a few little things.

### DIFF
--- a/asu/update.py
+++ b/asu/update.py
@@ -14,21 +14,24 @@ log = logging.getLogger("rq.worker")
 redis_client = get_redis_client()
 
 
-def update_targets(version):
+def update_targets(version) -> bool:
     """Update available targets of a specific version
 
     Args:
-        config (dict): Configuration
         version(str): Version within branch
     """
     branch_data = get_branch(version)
+    if "path" not in branch_data:
+        log.warning("Obsolete version %s, skipping...", version)
+        return False
+
     branch_name = branch_data["name"]
     version_path = branch_data["path"].format(version=version)
     targets = requests.get(settings.upstream_url + f"/{version_path}/.targets.json")
 
     if targets.status_code != 200:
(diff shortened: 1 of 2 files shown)


efahl:
I tried for a few hours to write a test for this going through the api.  Read this at least 10 times, https://fastapi.tiangolo.com/tutorial/testing/#extended-fastapi-app-file

I could never get the headers to pass through, the `x_update_token` value in the api update function was always `None`...  Everything else got through, and `token` inside `update` was `good`, as expected.

```
def test_api_update_good(app):
    settings.update_token = "good"
    client = TestClient(app)
    headers = {"X-Update-Token:": settings.update_token}
    response = client.get("/api/v1/update/21.02.7/x86/64", headers=headers)
    assert response.status == 204
```
In `asu/api.py:update`, these are the values we have during the test run.  Got any insight on how to get that header value passed in?
```
version = '21.02.7'
target = 'x86'
subtarget = '64'
response = <starlette.responses.Response object at 0x748b08442390>
x_update_token = None
settings.update_token = 'good'
```

efahl:
Got it!  Set the headers on the client and it works.
```
    settings.update_token = "good"
    client = TestClient(app)
    client.headers["X-Update-Token"] = settings.update_token

    response = client.get("/api/v1/update/21.02.7/x86/64")
    assert response.status_code == 204
```

PR #986, opened 2024-09-23
build: use exec_run with dmux

aparcar:
Since upstream implemented `dmux` for exec_run it should be used instead of creating containers multiple times. Previously it wasn't possible to tell stderr apart from stdout, making it harder to parse things. Now with proper demuxing, recycle the container all three times, yey!

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -21,7 +21,7 @@ from asu.util import (
     get_request_hash,
     parse_manifest,
     report_error,
-    run_container,
+    run_cmd,
 )
 
 log = logging.getLogger("rq.worker")
@@ -62,49 +62,6 @@ def build(build_request: BuildRequest, job=None):
     log.info(f"Pulling {image}...")
     podman.images.pull(image)
     log.info(f"Pulling {image}... done")
-
-    returncode, job.meta["stdout"], job.meta["stderr"] = run_container(
-        podman, image, ["make", "info"]
-    )
-
-    job.save_meta()
-
-    version_code = re.search('Current Revision: "(r.+)"', job.meta["stdout"]).group(1)
-
-    if requested := build_request.version_code:
-        if version_code != requested:
-            report_error(
-                job,
-                f"Received incorrect version {version_code} (requested {requested})",
-            )
-
-    default_packages = set(
-        re.search(r"Default Packages: (.*)\n", job.meta["stdout"]).group(1).split()
-    )
-    log.debug(f"Default packages: {default_packages}")
-
-    profile_packages = set(
-        re.search(
-            r"{}:\n    .+\n    Packages: (.*?)\n".format(build_request.profile),
-            job.meta["stdout"],
-            re.MULTILINE,
-        )
-        .group(1)
-        .split()
-    )
-
-    appy_package_changes(build_request)
-
-    build_cmd_packages = build_request.packages
-
-    if build_request.diff_packages:
-        build_cmd_packages: list[str] = diff_packages(
-            build_request.packages, default_packages | profile_packages
-        )
-        log.debug(f"Diffed packages: {build_cmd_packages}")
-
-    job.meta["imagebuilder_status"] = "calculate_packages_hash"
-    job.save_meta()
 
     mounts = []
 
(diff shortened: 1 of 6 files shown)


aparcar:
@efahl could you please have a look? This is a muuuch cleaner implementation (idea) compared to how it was done before. Error handling should be improved and `sleep 60` is awkward, however it should generally do the trick.

efahl:
Yes, much cleaner!  One question and one comment.

Question: what does the 'sleep 60' command do on the initial container create?  Just let all the mounts and stuff complete, before doing anything else?

Comment: would it be possible to put the signing mounts into the original container and eliminate creating the second one after the build?  I'm curious if that would save anything...

I installed podman==5.2.0, I assume that's the right version for all of this.

aparcar:
I added some more fixes, please test it. Also it's deployed here https://sysupgrade.staging.openwrt.org

aparcar:
> Yes, much cleaner! One question and one comment.

Yey!

> Question: what does the 'sleep 60' command do on the initial container create? Just let all the mounts and stuff complete, before doing anything else?

This could also be `sleep infinity`, point is that there needs to be a process that keeps running or else the container is considered stopped. Normally this would be a `nginx` process or whatever, but we only use docker for building.

> Comment: would it be possible to put the signing mounts into the original container and eliminate creating the second one after the build? I'm curious if that would save anything...

This is separated intentionally. Imagine someone sneaks in evil code to the packages.git feed which runs as post-install within the image builder. Separating the signing in a fresh container prevents an attacker from stealing signing keys. Worst case a malicious package is signed, however it's only delivered to whoever requests firmware with the malicious package anyway. tl;dr separate user input from signing keys via two containers.

> 
> I installed podman==5.2.0, I assume that's the right version for all of this.

Should start to work with 5.1.0

efahl:
I'm getting ''unable to find user buildbot: no matching entries in passwd file'' whenever we hit run_cmd.  Do I need to delete or regenerate something?

efahl:
Just did an `owut` build against the staging server, worked just fine:
```
23.05.4 x86/64...

Build: Requesting version r24012-d8dd03c46f (kernel 5.15.162)

Requesting build ----------------------
Downloaded https://sysupgrade.staging.openwrt.org/api/v1/build to /tmp/owut-build-response.json (144B at 0.001 Mbps)
Hash:   8f66be4d565089ded75797239e15ca76
...
Build completed in 77 seconds.
  build_at       = 2024-07-15T22:14:18.000000Z
  version_number = 23.05.4
  version_code   = r24012-d8dd03c46f (requested r24012-d8dd03c46f)
  rootfs_size_mb = default
  init-script    = no-init-script

Image source: https://sysupgrade.staging.openwrt.org/store/8f66be4d565089ded75797239e15ca76/openwrt-23.05.4-ae2949df9d71-x86-64-generic-ext4-combined-efi.img.gz
Downloaded https://sysupgrade.staging.openwrt.org/store/8f66be4d565089ded75797239e15ca76/openwrt-23.05.4-ae2949df9d71-x86-64-generic-ext4-combined-efi.img.gz to /tmp/firmware.bin (9666994B at 36.877 Mbps)
```

efahl:
> `unable to find user buildbot: no matching entries in passwd file`

I started up one of the testtarget-testsubtarget-v1.2.3 images manually, logged into it and the user on those is `builder` not `buildbot`, as it is in the newer image builder ones, hence all the test failures on my end.  If I set user='builder' in util.py, then all but the --runslow tests work (and vice versa).

So, what creates `ghcr.io/openwrt/imagebuilder:testtarget-testsubtarget-v1.2.3`?  Is that done manually?  I delete the image `podman rmi blah` and it just keeps coming back with a 3-month-old one.

aparcar:
I think it's an upstream issue but I'll also update the test container... https://github.com/containers/podman-py/pull/431

efahl:
> update the test container.

👍🏼 Works now!

aparcar:
I'd be curious to compare build times

efahl:
> I'd be curious to compare build times

That first test I ran above against your staging server at 77s was waaay faster than I've ever got from the production server, which is usually 180-190s.  Better hardware?

On my local server, I'm just running my `owut` regression suite, and the build time is about the same, 200s or so (I only give the VM 2 cores, so it's pretty slow).

efahl:
I rebuilt my local server with all the changes here, repopulated all the versions that I keep locally for testing and ran the `owut` suite against it.  No issues there.

Did some extra builds, to test both versions and speed:


| SNAPSHOT | x86/64 | generic |   130s |
| ---- | ---- | ---- | ---- |
| 23.05.4 | mediatek/mt7622 | linksys_e8450-ubi |   70s |
| 22.03.7 | ath79/generic | tplink,archer-c7-v4 |   70s |

I'd say it looks good!

aparcar:
Thanks for testing. x86 is notoriously slow, maybe @dangowrt got some ideas on how to improve that. I'll merge this once all tests pass after rebasing.

efahl:
Regarding timings, I did a few builds varying only one thing, how many extra packages were included.  With zero extra packages my snapshot-x86/64 build is ~75s.  With a couple dozen extra packages it goes up to ~180s.  It seems like those `opkg install` calls have the biggest impact on wall time, so not much to do there (and I'm guessing that it's probably not the `opkg` CPU use on ASU server, but more the network response when downloading the .ipk files that's the worst part).

Issue #866, opened 2024-06-25
"null" entry in package index


efahl:
There's a stray `"null": null` entry about halfway into one of the static package index files:
https://sysupgrade.openwrt.org/json/v1/snapshots/packages/x86_64-index.json
Current "modifed date" on the file is "6/25/2024, 5:20:22 AM" which would be PDT, so 12:20 UTC.

I first appeared in my `owut` test regressions at around 2024-06-24 about 20:00Z, possibly a couple hours earlier, but it's new and have not seen it before

Looks like it might be on the boundary between feeds, haven't looked at the asu code yet, just wanted to get a log while details are fresh in my mind.  Also have not look at other recently built platform package index files to see if it's present there.

```
"zyxel-reset": "2022.12.23~e1def7b5-r1", "null": null, "csstidy": "2021.06.13~707feaec-r1",
```

efahl:
Two hours later and the "null" entry is gone from x86, but still in aarch64.  I suspect this is an upstream glitch as I can't find anything in the asu update code that would generate the entry.

efahl:
"null" entry seems to be persisting in the aarch64 index files, but I don't see it elsewhere.  Last check 2024-06-27 12:30 PDT.

efahl:
"null" finally disappeared in the aarch64 builds, it was there yesterday.

Closing issue with no root cause ever discovered.

efahl:
Found root cause and fixed in https://github.com/openwrt/asu/pull/991

PR #991, opened 2024-09-25
util: improve package file parsing

efahl:
Occasionally, upstream package files are being terminated with extra newlines, resulting in '"null": null' entries in the resulting package index.  Strip the input text to avoid this.

Change the splitting method on the text to both reduce memory footprint and speed things up.

I reported this back in https://github.com/openwrt/asu/issues/866, but it was transient and I never got to the bottom of it.  Fixed now.

### DIFF
--- a/asu/util.py
+++ b/asu/util.py
@@ -316,23 +316,18 @@ def check_manifest(manifest, packages_versions):
 
 def parse_packages_versions(text: str) -> dict:
     index = {}
-    linebuffer = ""
     architecure = ""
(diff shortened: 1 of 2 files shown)


efahl:
Just discovered that adding `headersonly=True` to the parsestr call speeds the function up by about 10% overall, so put that in, too.
```
10 runs:
headersonly  :   1.812119s ( 88.8%)
original     :   2.039979s (100.0%)
```

PR #1001, opened 2024-09-29
api: clean up handling of mapped_profiles

efahl:
Replace special casing for x86 and armsr with more algorithmic approach. Fixes bug with x86/geode, which was silently transforming 'geos' profile into 'generic'.  Handles other generic platforms without code changes.

Also did general cleanup/refactoring of build validation error handling, and added more build validation test cases.

### DIFF
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -45,6 +45,11 @@ def api_v1_overview():
     return RedirectResponse("/json/v1/overview.json", status_code=301)
 
 
+def validation_failure(detail):
+    logging.info(f"Validation failure {detail = }")
+    return {"detail": detail, "status": 400}, 400
+
+
 def validate_request(build_request: BuildRequest):
     """Validate an image request and return found errors with status code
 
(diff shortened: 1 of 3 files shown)


asu/routers/api.py
```diff
@@ -60,80 +65,48 @@ def validate_request(build_request: BuildRequest):
...
+    if r.sismember(profiles_key, sanitized_profile):
+        logging.debug(f"Using {sanitized_profile = }")
+        build_request.profile = sanitized_profile
     else:
-        if r.sismember(
-            f"profiles:{branch}:{build_request.version}:{build_request.target}",
-            build_request.profile.replace(",", "_"),
-        ):
-            build_request.profile = build_request.profile.replace(",", "_")
+        mapped_profile = r.hget(f"mapping:{target_key}", build_request.profile)
```
aparcar, line 100: Where do you add those mapped profiles?

asu/routers/api.py
```diff
@@ -60,80 +65,48 @@ def validate_request(build_request: BuildRequest):
...
+        logging.debug(f"Checking {mapped_profile = }")
+        if mapped_profile:
+            build_request.profile = mapped_profile
+        elif r.sismember(profiles_key, "generic"):
```
aparcar, line 104: What do you think of this approach @dangowrt ? I wonder that if a profile name changes between releases, it would automatically select generic instead, may breaking the running system more than intended.

efahl: Hmm, yeah that would be bad.  Do we need a profile_changed table like package_changes?  (Maybe 
injects appropriate data into `mapping:`...)

Here are the `generic` profiles I found when I downloaded all of snapshot while doing preliminary work for this PR):
```
target: armsr/armv7          has profiles ['generic']
target: armsr/armv8          has profiles ['generic']
target: ath25/generic        has profiles ['generic', 'ubnt_picostation-2']
target: loongarch64/generic  has profiles ['generic']
target: octeon/generic       has profiles ['cisco_vedge1000', 'generic', 'itus_shield-router', 'ubnt_edgerouter', 'ubnt_edgerouter-4', 'ubnt_edgerouter-6p', 'ubnt_edgerouter-lite', 'ubnt_unifi-usg']
target: x86/64               has profiles ['generic']
target: x86/generic          has profiles ['generic']
target: x86/geode            has profiles ['generic', 'geos']
target: x86/legacy           has profiles ['generic']
```

Wasn't there a target change `mt78??` or something to `filogic` a while back, I wasn't paying attention but it seems to spark something in my memory.  How was that handled?  (As I recall, the `mtxx` name only appeared in snapshot, and then only for a short while???  So had to be done by user???)

I suppose the real solution would be make sure board.json contains a legit profile, make sure x86 and armsr both have `generic` when that's what it really is.  That or list the actual device name (one of my x86's is a `pc-engines-apu2`) in the profiles.json.

efahl: Looking at that list again, seems like maybe just also check if there's any choice in the matter:
```
    if len(target.profiles) == 1 and target.profiles[0] == "generic":
        # Assume "generic" as that's the only choice possible.
        profile = "generic"
```
I like this one.

efahl: Changed it to:
```
    elif r.scard(profiles_key) == 1 and r.sismember(profiles_key, "generic"):
```
Errors appropriately on wacky cases (geode, octeon, ath25), selects generic for all of the appropriate cases (x86, armsr, loongarch).

PR #1008, opened 2024-09-30
build: pass in correct VERSION_PATH for release snapshots

efahl:
Explicitly set the version path when creating the squid container so that we download the correct image builder for release snapshots.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -68,6 +68,9 @@ def build(build_request: BuildRequest, job=None):
         environment.update(
             {
                 "TARGET": build_request.target,
+                "VERSION_PATH": "snapshots"
+                if build_request.version == "SNAPSHOT"
+                else f"releases/{build_request.version}",
                 "use_proxy": "on",
                 "http_proxy": "http://127.0.0.1:3128",
             }


aparcar:
I wasn't sure it it might be lowercase, did you check that?

efahl:
> might be lowercase

Yes, upper case on downloads site, and at least `owut` makes sure the name in the build request is correct case (pretty sure `auc` and LuCI do, too).

https://downloads.openwrt.org/releases/23.05-SNAPSHOT/targets/

aparcar:
Oops now I just pushed this, sorry https://github.com/openwrt/asu/commit/b5490b72e32aab93683376641a7888f75107bd55

efahl:
> Oops now I just pushed this

Do you need me to do anything?  Or are you just going to wrestle with git for a while?

aparcar:
I mean I should be working. I think it's better to have it always lowercase in case some other client or web interface sends it any other way

efahl:
Yeah, the up/low case thing trips me up all the time.  Not sure why downloads requires upper case there.

image

PR #1025, opened 2024-10-08
util: set group:user on extracted tar members

efahl:
Set the gid:uid on the tar file members as we extract them from the build archive, so they are owned by a real user in both the container and on the host.

This came about from two things: part of documenting setup of a local server, and attempting to write host-based tools to manage build storage.

### DIFF
--- a/asu/util.py
+++ b/asu/util.py
@@ -4,6 +4,7 @@ import hashlib
 import json
 import logging
 import struct
+from os import getuid, getgid
 from pathlib import Path
 from re import match
 from tarfile import TarFile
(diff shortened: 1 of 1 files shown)


aparcar:
Did you check that this is not much slower?

efahl:
I didn't pay close attention, but the resulting builds were just as fast.  My ath79 live test using `owut` was the usual 30 seconds with and without.  Shouldn't cost much as it's just setting some bits in the archive member headers.

Live testing strategy:
1) Run all the usual pytests, when they're good:
2) Start up my local server using the new code.
3) Over on one of my routers, run `owut download --device <some test target>`
4) Check the total time, usually 30s on ath79 with no packages, 180s on x86 with a bunch of packages.
5) Back on the server, look at the asu_worker_1 logs and public/store dir to make sure everything looks good.

I usually do a few owut runs with and without `defaults`, add or remove packages, that sort of thing.

aparcar:
Great thanks for checking

aparcar:
Thanks for checking. Back in the day I terribly slowed down ASU by relying on pure python tar extraction

PR #1031, opened 2024-10-11
tests: add config to scan only tests directory

efahl:
Add a config entry to constrain pytest's search for test files and functions to only the 'tests/' directory.  This avoids error messages about read-only subdirectories, like you get when running squid.


efahl:
@aparcar  I had forgotten this one, it has that same strange transient error on Py 311 when checking the timestamp results...

PR #1033, opened 2024-10-13
api: add HEAD request handler for build status

efahl:
Allow clients to perform HEAD request for build status, significantly reducing the data sent during polling.  Does not change existing functionality in any way, so clients that do not send HEAD requests will work exactly as before.

Tested with a modified version of 'owut' and for long builds can save up to a megabyte of data transfer without any loss of function ('owut' simply reports current state during the build, so doesn't need anything more than the status code and the X-Imagebuilder-Status header value).

### DIFF
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -130,6 +130,7 @@ def return_job_v1(job):
             }
         )
         headers["X-Queue-Position"] = str(response["queue_position"])
+        headers["X-Imagebuilder-Status"] = "queued"
 
     elif job.is_started:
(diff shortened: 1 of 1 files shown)


efahl:
@aparcar, got any idea what's up with that `test_stats_image_builds` run that's failing in Py 3.11?

I think I've seen it before, but can't reproduce it locally.  I'm now running two test servers, one with Py 311 and the other on 312, both run that test successfully...

aparcar:
Not sure, a rerun fixed it

PR #1036, opened 2024-10-14
chore: improve typing and annotations on BuildRequest

efahl:
Add full annotations to all BuildRequest members, including the verbose descriptions from the pre-pydantic API.  This produces much better documentation in the redoc and api docs.  Also did a little semantic reordering, so the flow of the docs is a bit better.


aparcar:
Excellent thanks

PR #1051, opened 2024-10-23
tests: fix spurious failure of stats tests

efahl:
Some of the test_stats tests would fail occasionally with the wrong queue length.  Traced it back to the redis 'Queue' object somehow persisting across tests.  We now ensure that a fresh queue is used for each test.


efahl:
Second push includes the changes I made to test_stats.py to chase this down, specifically adds assertions to ensure that the queues are empty before doing anything else.

Also incorporates a bunch of cleanup to make the test cases much more readable.

PR #1064, opened 2024-10-29
Overwrite FileResponse Content-Type with application/octet-stream for all files in /store

### LINKED ISSUE #1059: Incorrect Content-Type results in firmware files opening in the browser
Download-Links to a notable amount of built Images via the Browser-Interface of the public [Firmware-Selector-Server](https://firmware-selector.openwrt.org/) on the [public ASU](https://sysupgrade.openwrt.org/) are being advertised as `content-type: text/stream; charset=utf-8` instead of from `content-type: application/octet-stream` the Webserver running on ASU.

This leads to the Issue where, when clicking on the Download-Button, instead of Firefox opening the Download-File-Dialog, it interprets the File as a Webpage and tries to render it. Workaround is clicking "Save Target as".

I am not sure if I should put this here or at firmware-selector-openwrt-org - therefore I created a [Ticket there too](https://github.com/openwrt/firmware-selector-openwrt-org/issues/53).

While Chrome Browser seems to take care of this misconfiguration, Firefox doesn't.

AliveDevil:
Fixes #1059

### DIFF
--- a/asu/fastapi/StaticFiles.py
+++ b/asu/fastapi/StaticFiles.py
@@ -1,0 +1,41 @@
+import os
+
+from fastapi.responses import FileResponse, Response
+from fastapi.staticfiles import StaticFiles as FastApiStaticFiles
+
+from starlette.staticfiles import PathLike
+from starlette.types import Scope
+
+class StaticFiles(FastApiStaticFiles):
+    def __init__(
+        self,
+        *,
+        directory: PathLike | None = None,
+        packages: list[str | tuple[str, str]] | None = None,
+        html: bool = False,
+        check_dir: bool = True,
+        follow_symlink: bool = False,
+    ) -> None:
+        super().__init__(
+            directory=directory, 
+            packages=packages, 
+            html=html, 
+            check_dir=check_dir, 
+            follow_symlink=follow_symlink
+        )
+    
+    def file_response(
+        self,
+        full_path: PathLike,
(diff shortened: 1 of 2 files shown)


AliveDevil:
Should these be squashed? @aparcar

tests/test_store.py
```diff
@@ -0,0 +1,36 @@
+from asu.config import settings
+from asu.fastapi.staticfiles import FileResponse
+
+store_path = settings.public_path / "store"
```
AliveDevil, line 4: This is ugly.

But conftest.py overwrites settings.public_path, unfortunately the main.py/app has already been constructed, with the previous old public_path.
Thus the conftest.py app()-setting of public_path to a temp-path doesn't work here.

aparcar:
Could you please try again and see if the issue is resolved? If so the Caddy docs are broken, caddy is broken or I misunderstand some stuff, however it worked on my test run with Firefox

0nelight:
A caddy team-member said:
"Your upstream app should set a correct Content-Type header.

https://caddy.community/t/caddy-reverse-proxy-sets-content-type/15604/2

This PR looks good from my point of view, but I have no development setup here so I didn't test it.

aparcar:
This is now fixed by letting Caddy server serving the `/store/` path. Caddy automatically determines the right header types and this issue should be gone.

PR #1076, opened 2024-11-08
config: enable use of new setup.sh in snapshots

efahl:
Run the setup.sh in the "new format" snapshot docker images prior to running any make commands.

@aparcar , probably need to do this soon as there are a lot of reports piling up on FS and elsewhere about "make info" not working, this is the fix.


aparcar:
Thanks, I solved It slightly differently since your PR would break any server not running a squid

https://github.com/openwrt/asu/commit/b7e3baab3fcd765806acc7e537518354f66bcd4e

efahl:
Excellent.  I've got local and remote tests running with your changes now, looks good!

PR #1080, opened 2024-11-12
util: read index.json for snapshot package feeds

efahl:
As of the apk switchover, the Packages file no longer is created as part of the package feeds.  Instead, utilize the index.json file created from apk's packages.db.

See https://github.com/openwrt/openwrt/commit/f841870


efahl:
Yeah, the ABI versioned names are currently present in the index.json files (which is why I restricted this to use index.json for snapshots, and Packages for all the "old" builds).  @Ansuel and I have discussed the ABI issue a bit on the forum, but it's an upstream issue and this gets the pipeline in place for the APK builds.  Once ABI-ed names issue gets resolved upstream, this should work without modification.

PR #1104, opened 2024-12-02
build: use new imagebuilder setup.sh for appropriate versions

efahl:
Properly detect versions for which the slim imagebuilder containers using setup.sh is being used.  Set up the container accordingly.

Allows building the new 24.10 versions, including the RCs.

### DIFF
--- a/asu/util.py
+++ b/asu/util.py
@@ -362,6 +362,11 @@ def parse_feeds_conf(url: str) -> list[str]:
     )
 
 
+def is_snapshot_build(version: str) -> bool:
+    """For imagebuilder containers using 'setup.sh' instead of fully populated."""
+    return version.lower().endswith("snapshot")
+
+
 def is_post_kmod_split_build(path: str) -> bool:
     """Root cause of what's going on here can be found at
     https://github.com/openwrt/buildbot/commit/a75ce026
@@ -381,7 +386,7 @@ def is_post_kmod_split_build(path: str) -> bool:
     if path.startswith("snapshots"):
         return True
 
-    version = path.split("/")[1]
+    version: str = path.split("/")[1]
     if version.startswith("24."):
         return True
     if version.startswith("23."):
--- a/asu/build.py
+++ b/asu/build.py
@@ -22,6 +22,7 @@ from asu.util import (
     get_podman,
     get_request_hash,
     parse_manifest,
+    is_snapshot_build,
     report_error,
     run_cmd,
 )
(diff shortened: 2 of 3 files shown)


efahl:
@aparcar you might want to get this one in, as it allows the ASU server to build the 24.10 RCs, which are broken without it...

asu/util.py
```diff
@@ -368,6 +368,21 @@ def parse_feeds_conf(url: str) -> list[str]:
     )
 
 
+def is_new_style_build(version: str) -> bool:
+    """For imagebuilder containers using 'setup.sh' instead of fully populated."""
+    if version.lower().endswith("snapshot"):
+        return True
+
+    if version.startswith("24."):
```
aparcar, line 376: The "new style" only applies to snapshots aka daily changing containers. Tagged versions don't change and contain the full image builder.

asu/util.py
```diff
@@ -368,6 +368,21 @@ def parse_feeds_conf(url: str) -> list[str]:
     )
 
 
+def is_new_style_build(version: str) -> bool:
```
aparcar, line 371: I think this naming is a bit.. unspecific. Isn't something like `is_snapshot` better for this?

asu/util.py
efahl: The rc builds are using it, too:

```
$ podman run -it ghcr.io/openwrt/imagebuilder:x86-64-v24.10.0-rc1
buildbot@d2ddfc54230f:~$ ls -la
total 28
drwxr-xr-x 3 buildbot buildbot 4096 Dec  2 10:35 .
dr-xr-xr-x 1 root     root     4096 Dec  3 17:01 ..
-rw-r--r-- 1 buildbot buildbot  220 Mar 27  2022 .bash_logout
-rw-r--r-- 1 buildbot buildbot 3526 Mar 27  2022 .bashrc
drwx------ 3 buildbot buildbot 4096 Dec  2 10:35 .gnupg
-rw-r--r-- 1 buildbot buildbot  807 Mar 27  2022 .profile
-rwxr-xr-x 1 root     root      925 Dec  2 10:35 setup.sh
```

efahl: There are actually two things changed, the kmod-split and the setup.sh-based containers.  Currently they overlap completely, with both applying to

1) All main SNAPSHOT and all release -SNAPSHOT builds.
2) All 24.10* and later.
3) As-yet unreleased 23.05.6 and later on that branch (I suppose any 22.x built after today, too).

What's the plan?  I suspect keep kmod-split as-is, but remove the setup.sh-based containers from all but SNAPSHOTs?

efahl: Is this the culprit for the setup.sh in the RC images?  

https://github.com/openwrt/docker/blob/main/Dockerfile#L26

Does it need some sort of logic like `ARG RUN_SETUP=!version.endswith("snapshot")` however you'd say that inside a docker file???

aparcar: Should be fixed now, thanks for letting me know

efahl: Will the already-existing rc2 images be replaced?  They got generated before you got the fix in, so I'm still seeing them with the bug.

In any case, I'll rework this PR to make everything correspond with the fix and push a new version.

efahl: Oops, unclear above: "rc2 images" should be "rc2 docker images on ghcr, quay etc."  The stuff on downloads is fine.

efahl:
Rebased.  Ends up not fixing anything, but the general cleanup and added tests are probably worth keeping.

PR #1140, opened 2024-12-24
# fix json_path ref public_path

czy21:


aparcar:
What situation do you fix with this?

czy21:
> What situation do you fix with this?

When use settings.json_path not get public_path value from .env

efahl:
I think this is now obsolete since `json_path` no longer exists (as there's no longer any json storage).

aparcar:
Oh right, thanks! Anyway thanks @czy21 for putting in the work!

PR #1155, opened 2025-01-17
build: do auto-replacement on language packages that have been renamed

efahl:
Utilize the 'apply_package_changes' mechanism to apply algorithmic replacement of language packages that have changed names at a major release boundary.

Also fixed the long-standing typo: 'appy' -> 'apply'

Links: https://forum.openwrt.org/t/owut-openwrt-upgrade-tool/200035/406

### DIFF
--- a/asu/package_changes.py
+++ b/asu/package_changes.py
@@ -5,7 +5,17 @@ from asu.build_request import BuildRequest
 log = logging.getLogger("rq.worker")
 
 
-def appy_package_changes(build_request: BuildRequest):
+# Language pack replacements are done generically on a per-version basis.
+# Note that the version comparison below applies to all versions the same
+# or newer, so for example "24.10" applies to snapshots, too.
+language_packs = {
+    "24.10": {
+        "luci-i18n-opkg-": "luci-i18n-package-manager-",
+    },
+}
+
+
+def apply_package_changes(build_request: BuildRequest):
     """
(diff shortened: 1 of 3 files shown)


asu/package_changes.py
```diff
@@ -42,3 +52,12 @@ def _add_if_missing(package):
...
             elif build_request.profile == "buffalo_wzr-hp-g300nh-rb":
                 _add_if_missing("kmod-switch-rtl8366rb")
+
+    # TODO: if we ever fully implement 'packages_versions', this needs rework
+    for version, packages in language_packs.items():
+        if build_request.version >= version:  # Includes snapshots
+            for i, package in enumerate(build_request.packages):
+                for old, new in packages.items():
+                    if package.startswith(old):
+                        lang = package.rsplit("-", 1)[1]
```
1715173329, line 62: this does not work if the language pack contains addtional `-`, for example luci-i18n-opkg-`zh-cn` will be evaluated as `cn`, then result the following error:
```
Collected errors:
 * opkg_install_cmd: Cannot install package luci-i18n-package-manager-cn.
 *
```

aparcar: @efahl can you please have a look?

1715173329: Thinking a simple fix could be `lang = package.replace(old, "")`

PR #1163, opened 2025-01-24
RFC api: when specific build request failures occur attempt to recover

efahl:
If a build request has been received for a version, branch, target or subtarget that somehow has missed an update via webhooks, we invoke the 'update' mechanism and proceed with the error processing.

The result is that this first build request still fails, but any future requests on the branch/target will (hopefully) succeed.

---

I put this together as an experiment about the time rc2 came out, and have just been letting it run on my local test-only server, so it's seen a bunch of use during `owut` testing.

It works well with that one exception noted in the commit message: first person to do a build after the webhook failure gets a 400 `Unsupported profile: ...` error.  This isn't terrible for  `owut` users, because `owut` says "try again later", but LuCI app and FS just give uninformative messages so users are left wondering what is going on (maybe the failure messages produced by `validation_failure` should be augmented with the "try again later").

If we could re-queue the build request somehow _after_ the update request, that would make this much better, but the problem is saying "only try this once", as we don't want an infinite loop if the update attempt fails again.

### DIFF
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -51,6 +51,18 @@ def validation_failure(detail: str) -> tuple[dict, int]:
     return {"detail": detail, "status": 400}, 400
 
 
+def _attempt_update(build_request: BuildRequest) -> None:
+    """If the profile has not been updated recently, we may be able to heal
+    things by forcing an update.
+    """
+    get_queue().enqueue(
+        update,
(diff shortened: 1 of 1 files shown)


aparcar:
With the latest cache rework this is no longer required, right?

efahl:
Correct!

PR #1164, opened 2025-01-24
No hook but cache

aparcar:
Stop relying on web hooks but instead dynamically update whatever is required.


asu/routers/api.py
```diff
@@ -88,28 +94,42 @@ def validate_request(build_request: BuildRequest) -> tuple[dict, int]:
...
-        if mapped_profile:
-            build_request.profile = mapped_profile
-        elif r.scard(profiles_key) == 1 and r.sismember(profiles_key, "generic"):
-            # Assume "generic" as that's the only choice possible.
-            # Handles the x86, armsr and other generic variants.
-            logging.info(f"Use generic profile replacing {build_request.profile = }")
-            build_request.profile = "generic"
-        else:
+
+    if (
```
efahl, line 104: Assuming that we don't have a case where `reload` has picked up only `generic` when there are other variants for a given platform, this should work.  It seems pretty unlikely that we'd have that happen and the only case where it can is in x86/geode, where "generic" and "geos" are both present.  All of the rest of armsr, loongarch and x86 targets only have that single profile "generic".

Here's what I had that old code doing (completely untested):
```
def valid_profile(profile: str, build_request: BuildRequest) -> bool:
    profiles = app.profiles[build_request.version][build_request.target]
    if profile in profiles:
        return True
    if len(profiles) == 1 and "generic" in profiles:
        # Handles the x86, armsr and other generic variants.
        build_request.profile = "generic"
        return True
    return False
```
Then that code below turns into
```
    sanitized_profile = build_request.profile.replace(",", "_")

    if not valid_profile(sanitized_profile, build_request):
        reload_profiles(app, build_request.version, build_request.target)
        if not valid_profile(sanitized_profile, build_request):
            return validation_failure(f"Unsupported profile: {build_request.profile}")

```

efahl:
Except for that part you pointed out, looks ready to go.  

Are you going to roll it out to a test instance first?  Let me know the url if you do.

efahl:
We've got maybe 20 tests against the staging server, across a variety of targets and versions.  https://forum.openwrt.org/t/owut-openwrt-upgrade-tool/200035/432

Looks good!

efahl:
@aparcar I assume you've seen the 500s from various calls by now.  Here are the headers from one of them, in case it has anything useful:

```
$ curl -v https://sysupgrade.staging.openwrt.org/json/v1/releases/24.10-SNAPSHOT/targets/x86/64/index.json
> GET /json/v1/releases/24.10-SNAPSHOT/targets/x86/64/index.json HTTP/2
> Host: sysupgrade.staging.openwrt.org
> User-Agent: curl/8.10.1
> Accept: */*
>
< HTTP/2 500
< access-control-allow-credentials: true
< access-control-allow-headers: *
< access-control-allow-methods: POST, GET, OPTIONS
< access-control-allow-origin: *
< access-control-expose-headers: Authorization
< alt-svc: h3=":443"; ma=2592000
< content-type: text/plain; charset=utf-8
< date: Wed, 29 Jan 2025 02:15:39 GMT
< server: Caddy
< server: uvicorn
< vary: Origin
< content-length: 21
<
Internal Server Error
```

PR #1188, opened 2025-02-03
build: ensure request hash is the same as build hash

efahl:
When using 'packages_versions' the request hash that is calculated before request validation in 'api_v1_build_post' often differs from the one calculated in 'build'.  This is due to 'validate_request' modifying the request's package lists.

We extract the request hash from the job, thus making them consistent again.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import Union
 
 from rq import get_current_job
+from rq.job import Job
 
 from asu.build_request import BuildRequest
 from asu.config import settings
@@ -20,7 +21,6 @@ from asu.util import (
     get_container_version_tag,
     get_packages_hash,
     get_podman,
-    get_request_hash,
     is_snapshot_build,
     parse_manifest,
     report_error,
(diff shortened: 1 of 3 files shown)


aparcar:
Thanks, good point. Isn't something like below the "better" solution?

```
diff --git a/asu/util.py b/asu/util.py
index 0bf1b87..8704b74 100644
--- a/asu/util.py
+++ b/asu/util.py
@@ -141,7 +141,12 @@ def get_request_hash(build_request: BuildRequest) -> str:
                 build_request.version_code,
                 build_request.target,
                 build_request.profile.replace(",", "_"),
-                get_packages_hash(build_request.packages),
+                get_packages_hash(
+                    x.removeprefix("+")
+                    for x in (
+                        build_request.packages_versions.keys() or build_request.packages
+                    )
+                ),
                 get_manifest_hash(build_request.packages_versions),
                 str(build_request.diff_packages),
                 "",  # build_request.filesystem
```

aparcar:
How are the results more reproducible? I think the computation of the hash is negligible, I don't know what the "get current job" function does in the background.

efahl:
Reproducible in the sense that you can call `get_request_hash` before or after the validation step and you get the same result.

Looks like `get_current_job` is how we get `job` in the build function.  The `job` parameter is completely ignored, never used by `rq` at all so we can get rid of it altogether.

I see no way of passing the actual job into build, since it's created after the call to `enqueue` and not included in the call (just the `*args` and `**kwds` are passed to the queued function, meaning we could actually pass the request hash in as an argument; I'll mangle this PR and you can take a look at it).

https://github.com/rq/rq/blob/master/rq/queue.py#L965
https://python-rq.org/docs/jobs/#job-creation

PR #1199, opened 2025-02-06
fix: replace `auc` with `owut` for 24.10 releases

aparcar:
The cleaner implementation is to handle this on the clients. @efahl do you work on the LuCI app? If not we should consider dropping it or find someone else to maintain it.


efahl:
Yes, I'll see about adding the client code to LuCI.

As far as I can tell, that's the biggest missing piece in the LuCI app for cross-release upgrades, it doesn't do the `package_changes` processing.  Do you have any other things that you'd like to see in it?

PR #1221, opened 2025-02-08
build: apply-package-changes: handle long language identifiers

efahl:
Fix a bug in language packages name updates when the language identifier contains a dash.  Examples are "zn-ch" or "zh-hant".

Reported-by: @1715173329
Links: https://github.com/openwrt/asu/pull/1155#discussion_r1947690063


1715173329:
If possible please use this tag instead:
```
Reported-by: Tianling Shen <cnsztl@immortalwrt.org>
```

I don't personally suggest using @ in commit (message), as it could lead annoying notification email, especially for a popular project where forks and cherry-pick are happening.

PR #1222, opened 2025-02-08
update: append upcoming version to reload versions

1715173329:
Add missing upcoming version to reload versions.


1715173329:
Oh looks like it already applied, sorry for the extra noise.

aparcar:
Well feel free to suggest more changes. Also I'd be nice if you could ask the immortalwrt community if they could help out with the LuCI app, it's been improvised for a long time since no real JS developer worked on it

PR #1223, opened 2025-02-08
fix: remove the container once it has failed

a-gave:
This is a minor fix that removes failed containers, before the report_error function, to free up what little space they would occupy in ram until the timeout.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -175,6 +175,7 @@ def build(build_request: BuildRequest, job=None):
             container, ["sh", "setup.sh"]
         )
         if returncode:
+            container.kill()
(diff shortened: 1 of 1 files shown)


aparcar:
Lovely, thanks!

aparcar:
Another nice feature would be to remove those leftover volumes/images etc, they somewhat fill up the storage and I run a cronjob like `podman system prune -af`.

a-gave:
Not yet tested but I guess changing the last container.stop(), and also these container.kill(), with the equal of 'podman rm -vf' should do the job https://podman-py.readthedocs.io/en/latest/_modules/podman/domain/containers.html#Container.remove

PR #1254, opened 2025-02-15
fix: keep request_hashe consistent

aparcar:


asu/util.py
```diff
@@ -164,9 +166,19 @@ def get_packages_hash(packages: list[str]) -> str:
...
 
     Returns:
-        str: hash of `req`
+        str: hash of `packages`
     """
-    return get_str_hash(" ".join(sorted(list(set(packages)))))
+    return get_str_hash(
+        " ".join(
+            sorted(
+                list(
```
efahl, line 174: You can get rid of the `list()` as `sorted` automatically converts the set to a list...
```
    return get_str_hash(" ".join(sorted(set(x.removeprefix("+") for x in packages))))
```

efahl:
Tested and works as expected.

PR #1274, opened 2025-02-25
feat: improvements in package caching

a-gave:
This adds some squid directives to consider fresh longer kernel modules of already compiled releases, to prevent revalidation against the origin server which consumes time.

An automatic mechanism to refresh the below squid directive should be included (e.g., 24h after the first appearance on https://downloads.openwrt.org/.versions.json)
```
refresh_pattern ^http://downloads.openwrt.org/releases/.*(23.05.[6-99]|24.10.[1-99]|25.*|SNAPSHOT)/targets/.* 0 100% 0 refresh-ims
```

This adds also a fallback to no_proxy (https) in case STDERR contains “package index are corrupt” to rule out cache problems and to allow packages to be considered fresh longer and thus not revalidated against the origin server

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -222,21 +222,37 @@ def build(build_request: BuildRequest, job=None):
     job.meta["imagebuilder_status"] = "validate_manifest"
     job.save_meta()
 
+    job.meta["make_manifest_cmd"] = [
+        "make",
+        "manifest",
+        f"PROFILE={build_request.profile}",
+        f"PACKAGES={' '.join(build_cmd_packages)}",
+        "STRIP_ABI=1",
+    ]
+
     if settings.squid_cache and not is_snapshot_build(build_request.version):
         log.info("Disabling HTTPS for repositories")
         # Once APK is used for a stable release, handle `repositories`, too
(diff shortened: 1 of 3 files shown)


aparcar:
I recon you're really limited on bandwidth? Or what is the motivation to avoid the extra checks?

a-gave:
Yes, this is purely to save bandwidth and to give a crutch to somewhat dated processors.

A build without selecting additional packages, just changing defaults

With a recent processor:
1- 43 seconds (refreshing everything)
2- 39 seconds (refreshing only packages, not kmods)
3- 22 seconds (without any refresh and with all TCP_MEM_HIT from ram)

The same with a 10years old processor:
1- 48s
2- 42s
3- 31s

Also the regex below should cause less fallback to https, if building with feeds from a mirror 
(that resynchronizes i.e. every 12 or 24 hours)
```
refresh_pattern ^http://.*openwrt.*/.* 30 100% 30 override-expire
```
Close it if not necessary, maybe it could be useful as a reference for someone else, but maybe not.

a-gave:
Obsoleted by https://github.com/openwrt/asu/commit/331c976238e7c72d5d6354827a6491bde812d3b5 (squid replaced by nginx-cache)

PR #1280, opened 2025-03-01
config: remove redundant entry in package_changes

efahl:
We somehow got a duplicate entry for the libustream change in the table, remove it.


efahl:
@aparcar , ping...  Sometime when you're doing housekeeping, this one should go in.

PR #1361, opened 2025-04-18
Changing publish action for multi arch

RodrigoPrestes:
This PR mainly aims to add support for Docker/Podman on arm64, currently [the latest tag](https://hub.docker.com/r/openwrt/asu/tags) only supports amd64.
Following the [README for build-push-action](https://github.com/docker/build-push-action?tab=readme-ov-file#usage) the change should straightforward by just adding two steps.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -54,8 +54,14 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
+      - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


aparcar:
Hi @RodrigoPrestes thanks for your work. I'm afraid this won't work since all ImageBuilders are compiled for x86 only, or do you plan a distributed infrastructure where builds happen on dedicated (x86) servers?

RodrigoPrestes:
Since pip and poetry works on arm I thought it was just a question of building for it.
If there's more work to accomplish this then I might not be the right person for that.
Closing the MR.

PR #1363, opened 2025-04-18
utils: eliminate cache races for .versions.json and profiles.json

efahl:
There were two code paths reading .versions.json, the first of which, in main:generate_latest, would update the cache version appropriately, but then when utils:reload_versions was called, it would see from_cache was true and never update its internal data structures.  The code paths were consolidated, so there is only one touch point for the .versions.json file now, eliminating the race.

Likewise, the version/target/profiles.json files are being read on many paths, but util:reload_profiles was basing its decision to update on the same flawed from_cache check.  Eliminating the cache check altogether seemed most appropriate here, as there's no easy way to consolidated the various uses of the profile.json file into a single point of contact.

### DIFF
--- a/asu/main.py
+++ b/asu/main.py
@@ -37,6 +37,7 @@ app.mount("/static", StaticFiles(directory=base_path / "static"), name="static")
 
 templates = Jinja2Templates(directory=base_path / "templates")
 
+app.latest = []
 app.versions = []
 reload_versions(app)
 logging.info(f"Found {len(app.versions)} versions")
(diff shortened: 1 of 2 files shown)


efahl:
@dc366 @systemcrash

Could I bother you two for a sanity check on this?  No need for deep code analysis, but if you could read through the commit message and see if it makes sense to you, that would be great.

systemcrash:
Decent. Commit message: `to consolidated` -> `to consolidate`

systemcrash (approved):

dc366:
I understand the issue and the message makes sense, glad you are fixing this.

PR #1369, opened 2025-04-20
build: put a limit on the number of queued build requests

efahl:
Restrict the number of build requests that may be in the queue by returning status=529 (server overloaded).

From observing the rate at which jobs are processed on the production server, it appears they are cleared at an average of just over 5 jobs/minute, so the selected queue length limit of 200 would equate to about a 40 minute wait.

### DIFF
--- a/asu/config.py
+++ b/asu/config.py
@@ -71,6 +71,7 @@ class Settings(BaseSettings):
     server_stats: str = ""
     log_level: str = "INFO"
     squid_cache: bool = False
+    max_pending_jobs: int = 200
 
 
 settings = Settings()
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -230,6 +230,14 @@ def api_v1_build_post(
             response.status_code = status
             return content
 
+        job_queue_length = len(get_queue())
+        if job_queue_length > settings.max_pending_jobs:
+            return {
+                "status": 529,  # "Site is overloaded"
+                "title": "Server overloaded",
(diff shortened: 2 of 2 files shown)


efahl:
Here's how `owut` responds:
```bash
$ owut upgrade
owut - OpenWrt Upgrade Tool %%VERSION%% (/root/bin/owut)
ASU-Server     http://asu-prod.lan:8000
...
Request:
  Version 24.10-SNAPSHOT r28601-9e9f6875bc (kernel 6.6.86)
--
Status:   server overload, queue contains too many build requests: 201
Progress:   0s total =   0s in queue +   0s in build

Build failed in   0s total =   0s in queue +   0s to build:
ERROR: Build failed with status 529
```

aparcar:
I think that's a bit of a work around. I think the better solution is to limit cached builds to 1 hour and thereby keep the required storage to a minimum. Right now we run out of storage during peak usage

efahl:
> Right now we run out of storage during peak usage

I made a garbage collector a while back by subclassing `rq.Worker` and then hooking it into the system via the standard worker class mechanism: 
```
    command: rqworker --logging_level INFO --with-scheduler --worker-class asu.GCWorker
```

It runs every 600 seconds by default (can be changed with `--maintenance-interval <secs>`) and it scans for expired jobs, then does `rm -rf` on job's `public/store/xxx` directory.  I'll make a PR and you can take a look.

efahl:
Resolved in https://github.com/openwrt/asu/commit/ac978d6d4da7adbe5b32703f04c2d313c34c4f5f

PR #1370, opened 2025-04-20
rq: add a garbage collector to the worker

efahl:
Implement a maintenance hook on the standard redis queue worker to do garbage collection on expired builds.  When a result expires from the queue, its data will be removed from the public/store/ directory at the regular maintenance interval (default is every 600 seconds).


efahl:
I've been running this for about 9 months now, no issues on my local server.

I test it out by setting the `ttl` values way down in `api.py` and watching the logs on the server after I queue up a bunch of builds.
```
+    failure_ttl: str = "3m"
+    result_ttl: str = "15m"
```

efahl:
But, are we running out of space due to old podman/docker containers hanging around, too?

.github/workflows/publish.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Set __version__ and poetry version
         run: |
           TAG="$(git describe --tags --always | awk -F"-" '{if (NF>1) {print substr($1, 2)".post"$2} else {print substr($1, 2)}}')"
-          echo "__version__ = \"$TAG\"" > asu/__init__.py
+          sed "s/__version__.*/__version__ = \"$TAG\"/" -i asu/__init__.py
```
aparcar, line 32: this change seems unrelated?

efahl: It's so the other contents of the __init__ file are not deleted.

efahl:
Just been playing with the podman API, this might be added to that worker, too.  It appears that pruning only affects unused containers/images, so you can safely just call these and reclaim the orphaned space.

```
$ python -c 'from asu.util import get_podman; print(get_podman().containers.prune())'
{'ContainersDeleted': [], 'SpaceReclaimed': 0}

$ python -c 'from asu.util import get_podman; print(get_podman().images.prune())'
{'ImagesDeleted': [{'Deleted': '4a2e6bf10ec5d4c70ff5b8598193285b6d090d47ad737bfe322c0be298e1e6d1', 'Untagged': ''}], 'SpaceReclaimed': 927715421}
```

aparcar:
Right now this is my crontab, however we could also use the GC worker:

```
@daily cd /srv && find ./store  -ctime +1 -exec rm -rf {} \;
@daily podman volume prune -af
```

efahl:
Obsoleted by recent additions, mostly https://github.com/openwrt/asu/commit/874acadcfd9e9674f27e81e97011ed242031582a

PR #1432, opened 2025-06-20
build: handle errors during container download

efahl:
If a container image is missing for whatever reason, then the image pull raises an exception:

 ImageNotFound: 404 Client Error: Not Found ({"message":"manifest unknown"}

This is propagated to the ASU clients as a "status 500 - Internal server error".  The ASU clients then simply report "init" as an error with no supporting details, leaving the user completely in the dark as to what is going on.

We add an error handler around the image pull, so now the server reports very specific details about the failure, helping diagnose the issue:

 RuntimeError: Image not found: ghcr.io/openwrt/imagebuilder:lantiq-xrx200-v24.10.1

Link: https://forum.openwrt.org/t/luci-attended-sysupgrade-support-thread/230552/68

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -8,6 +8,7 @@ from typing import Union
 from time import perf_counter
 
 from rq import get_current_job
+from podman import errors
 
 from asu.build_request import BuildRequest
 from asu.config import settings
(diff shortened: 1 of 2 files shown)


efahl:
I just noticed that this addresses https://github.com/openwrt/firmware-selector-openwrt-org/issues/99

PR #1455, opened 2025-08-03
config: document procedure for updating package_changes

efahl:
Provide a check list for adding entries to the 'package_changes' table.  Specifically describe how to calculate the proper revision number for a new entry.


aparcar:
Thanks!

PR #1462, opened 2025-08-14
fix: set correct media type for firmware

aparcar:
This shifts filetype handling from reverse proxy to client, allowing smaller setups.

Previously some files would be served as text, resulting i.e. Firefox browser to flood the screen with binary text instead of downloading it.

Fix: #1059


erik434:
Thanks @aparcar!! Much appreciated; confirms this fixes the problem without needing any mime.types file or reverse proxy, etc.

PR #1466, opened 2025-08-16
util: handle missing .targets.json file

efahl:
During the creation of a new maintenance release, it is possible that the new version's .target.json file has not yet been created. This causes a "500 internal server" error for all users, even if they request a different version.  Rather than crash attempting to decode the 404 html code as json, we simply return an empty dictionary.

This results in users seeing the "Unsupported target" message if they try to build the new version, but that's better than crashing for everyone.

### DIFF
--- a/asu/util.py
+++ b/asu/util.py
@@ -517,7 +517,7 @@ def reload_targets(app: FastAPI, version: str) -> bool:
     if response.extensions["from_cache"] and app.targets[version]:
         return False
 
-    app.targets[version] = response.json()
+    app.targets[version] = response.json() if response.status_code == 200 else {}
 
     return True
 


efahl:
@aparcar  You might want to expedite this one, as everything is broken with "500 internal server error" until the 23.05.6 targets.json is created...

```
$ owut check --version-to 24.10.2
owut - OpenWrt Upgrade Tool %%VERSION%% (/root/bin/owut)
ERROR: Response status 500 while downloading
  https://sysupgrade.openwrt.org/json/v1/overview.json
```
because of
```
$ curl https://downloads.openwrt.org/releases/23.05.6/.targets.json
<html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

efahl:
@aparcar 
👍🏼 This closes https://github.com/openwrt/asu/issues/1467

PR #1511, opened 2025-10-10
build-request: sanitize profile before use

efahl:
Older versions of the LuCI ASU app did not sanitize the device profile by replacing commas with underscores, it simply inserted the raw output from 'ubus call system board' into the request. This causes the build request to be rejected by the ASU server:

    Unsupported profile: sinovoip,bananapi-m2-berry. The requested
        profile was either dropped or never existed...

Let's make sure to sanitize profiles as soon as we receive them, to avoid this issue and ensure that all of our use is consistent with the upstream profile lists.

Link: https://forum.openwrt.org/t/luci-attended-sysupgrade-support-thread/230552/246

### DIFF
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -211,6 +211,9 @@ def api_v1_build_post(
     request: Request,
     user_agent: str = Header(None),
 ):
+    # Sanitize the profile in case the client did not (bug in older LuCI app).
+    build_request.profile = build_request.profile.replace(",", "_")
+
     request_hash: str = get_request_hash(build_request)
     job: Job = get_queue().fetch_job(request_hash)
     status: int = 200
--- a/asu/util.py
+++ b/asu/util.py
@@ -141,7 +141,7 @@ def get_request_hash(build_request: BuildRequest) -> str:
                 build_request.version,
                 build_request.version_code,
                 build_request.target,
-                build_request.profile.replace(",", "_"),
+                build_request.profile,
                 get_packages_hash(
                     build_request.packages_versions.keys() or build_request.packages
                 ),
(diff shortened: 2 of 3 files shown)


map-b:
Since the root cause of the `bananapi-m2-berry` case is that the dts compatible ("`sinovoip,bpi-m2-berry`" https://elixir.bootlin.com/linux/v6.12.57/source/arch/arm/boot/dts/allwinner/sun8i-v40-bananapi-m2-berry.dts#L51) neither it does not match the `makefile device name` (`Device/sinovoip_bananapi-m2-berry`) and it is not in `SUPPORTED_DEVICES ` (`SUPPORTED_DEVICES:=lemaker,bananapi-m2-berry`) was expected to fail. And that specific string `sinovoip,bananapi-m2-berry` comes from a suggestion to try to workaround the issue 

> Run echo 'sinovoip,bananapi-m2-berry' >| /tmp/sysinfo/board_name to change it to the new name.
https://forum.openwrt.org/t/luci-attended-sysupgrade-support-thread/230552/238

This misleaded to this "fix" and this resulted in some unexpected consecuencies. The worst, in my opinion, is that this will disalllow to have several devices profiles with SUPPORTED_DEVICES in common avoiding its support with ASU's clients https://github.com/openwrt/asu/issues/1525#issuecomment-3494315716.

To avoid totally hide this subtle "regression" I proposed to revert this PR commit https://github.com/openwrt/asu/commit/2e9edfd62ed5b079348b1e27c8f19024603d0420 and the posterior fix https://github.com/openwrt/asu/commit/1579236ca291a13b1f9f8b2e2552a776821e642d.

PR #1529, opened 2025-10-31
build: eliminate kill of already stopped containers

efahl:
Some error paths result in the container.kill being called twice. This results in an erroneous error report to the user, with the podman error superseding the actual build error in the message output:

    409 Client Error: Conflict (can only kill running containers...

Fix it by removing the extra calls.


efahl:
See https://github.com/openwrt/asu/issues/1530 as an example...

efahl:
@aparcar 
Can we get this bug fix in, too?  It's causing me headaches with trying to figure out root cause of issues due to it masking real errors.

aparcar:
Deployed

Issue #1525, opened 2025-10-22
Attended Sysupgrade OpenWrt 24.10.3 to 24.10.4 on GL-MT2500


alramon:
```
root@OpenWrt:~# owut upgrade --verbose
owut - OpenWrt Upgrade Tool 2025.09.27~f049043e-r1 (/usr/bin/owut)
ASU-Server     https://sysupgrade.openwrt.org
Upstream       https://downloads.openwrt.org
Target         mediatek/filogic
Profile        glinet_gl-mt2500
Package-arch   aarch64_cortex-a53
Root-FS-type   squashfs
Sys-type       sysupgrade
Version-from   24.10.3 r28872-daca7c049b (kernel 6.6.104)
Version-to     24.10.4 r28959-29397011cc (kernel 6.6.110)
Build-commit   https://git.openwrt.org/?p=openwrt/openwrt.git;a=shortlog;h=29397011cc
Build-FS-type  squashfs
Build-at       2025-10-19T16:37:45Z (~3 days ago)
Image-prefix   openwrt-24.10.4-mediatek-filogic-glinet_gl-mt2500
Image-URL      https://downloads.openwrt.org/releases/24.10.4/targets/mediatek/filogic
Image-file     openwrt-24.10.4-mediatek-filogic-glinet_gl-mt2500-squashfs-sysupgrade.bin
Installed      168 packages
Top-level       52 packages
Default         35 packages
User-installed  23 packages (top-level only)

Package version changes:
  base-files                          1662~daca7c049b                             1665~29397011cc
  kmod-crypto-acompress               6.6.104-r1                                  6.6.110-r1
  kmod-crypto-aead                    6.6.104-r1                                  6.6.110-r1
  kmod-crypto-authenc                 6.6.104-r1                                  6.6.110-r1
  kmod-crypto-crc32c                  6.6.104-r1                                  6.6.110-r1
  kmod-crypto-des                     6.6.104-r1                                  6.6.110-r1
  kmod-crypto-hash                    6.6.104-r1                                  6.6.110-r1
  kmod-crypto-hmac                    6.6.104-r1                                  6.6.110-r1
  kmod-crypto-hw-safexcel             6.6.104-r1                                  6.6.110-r1
  kmod-crypto-kpp                     6.6.104-r1                                  6.6.110-r1
  kmod-crypto-lib-chacha20            6.6.104-r1                                  6.6.110-r1
[...]

PR #1554, opened 2025-12-26
asu: revert sanitize 

map-b:
Since it seems there is no enough review time https://github.com/openwrt/asu/pull/1533 and the PR fixing the underlying issue which misleaded to the referenced revert commits has been merged:
- https://github.com/openwrt/openwrt/pull/21095

maybe just reverting to an known working state have better luck.

This would allow to use firmware-selector for the gl-mt2500 MaxLinear PHY case even Owut or Luci ASU app with the workaround of `echo "glinet_gl-mt2500" > /tmp/sysinfo/board_name` for the MaxLinear users.

The key here, IMHO, is that we have to make clearer the difference between "profile" and "DT compatible string" and for that I am thinking about some variable/messages renames could help also, but in order to prioritize, the code implied in the proposed reverts is the key and I think it goes in opposite direction.
https://github.com/openwrt/asu/issues/1525#issuecomment-3497242570
Right now we got this when generating the app.profiles[version][target]
```
{
'glinet_gl-mt2500': 'glinet_gl-mt2500',
'glinet_gl-mt2500-emmc': 'glinet_gl-mt2500',
'glinet_gl-mt2500-airoha': 'glinet_gl-mt2500',
'glinet_gl-mt2500': 'glinet_gl-mt2500',
'glinet_gl-mt2500-airoha': 'glinet_gl-mt2500-airoha',
'glinet_gl-mt2500-emmc': 'glinet_gl-mt2500-airoha',
'glinet_gl-mt2500': 'glinet_gl-mt2500-airoha',
'glinet_gl-mt2500-airoha': 'glinet_gl-mt2500-airoha'
}
which ends up like this: (not working)
{
'glinet_gl-mt2500': 'glinet_gl-mt2500-airoha',
'glinet_mt2500-emmc': 'glinet_gl-mt2500-airoha',
'glinet_gl-mt2500-airoha': 'glinet_gl-mt2500-airoha'
}
```
Before commit "sanitize profile" https://github.com/openwrt/asu/commit/1579236ca291a13b1f9f8b2e2552a776821e642d
```
{
'glinet,gl-mt2500': 'glinet_gl-mt2500',
'glinet,gl-mt2500-emmc': 'glinet_gl-mt2500',
'glinet,gl-mt2500-airoha': 'glinet_gl-mt2500',
'glinet_gl-mt2500': 'glinet_gl-mt2500',
'glinet,gl-mt2500-airoha': 'glinet_gl-mt2500-airoha',
'glinet,gl-mt2500-emmc': 'glinet_gl-mt2500-airoha',
'glinet,gl-mt2500': 'glinet_gl-mt2500-airoha',
'glinet_gl-mt2500-airoha': 'glinet_gl-mt2500-airoha'
}
which ends up like this: (working only works firmware-selector which use the profile directly, no DTS compatible)
{
'glinet,gl-mt2500': 'glinet_gl-mt2500-airoha',
'glinet,mt2500-emmc': 'glinet_gl-mt2500-airoha',
'glinet,gl-mt2500-airoha': 'glinet_gl-mt2500-airoha',
'glinet_gl-mt2500': 'glinet_gl-mt2500',
'glinet_gl-mt2500-airoha': 'glinet_gl-mt2500-airoha'
}
```

This do not implies nothing about which of PR fixing the gl-mt2500 issue in the openwrt repo will be finally merged:
- https://github.com/openwrt/openwrt/pull/20632 or
- https://github.com/openwrt/openwrt/pull/20947

### DIFF
--- a/asu/routers/api.py
+++ b/asu/routers/api.py
@@ -209,9 +209,6 @@ def api_v1_build_post(
     request: Request,
     user_agent: str = Header(None),
 ):
-    # Sanitize the profile in case the client did not (bug in older LuCI app).
-    build_request.profile = build_request.profile.replace(",", "_")
-
     add_build_event("requests")
 
     request_hash: str = get_request_hash(build_request)
--- a/asu/util.py
+++ b/asu/util.py
@@ -168,7 +168,7 @@ def get_request_hash(build_request: BuildRequest) -> str:
                 build_request.version,
                 build_request.version_code,
                 build_request.target,
-                build_request.profile,
+                build_request.profile.replace(",", "_"),
                 get_packages_hash(
                     build_request.packages_versions.keys() or build_request.packages
                 ),
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -537,7 +537,7 @@ def test_api_build_real_ath79(app):
             target="ath79/generic",
             version="23.05.5",
             packages=["tmux", "vim"],
-            profile="8dev,carambola2",  # Test unsanitized profile.
+            profile="8dev_carambola2",
         ),
     )
 


aparcar:
@efahl what's your opinion here?

efahl:
Completely unneeded as the root cause has been fixed upstream in https://github.com/openwrt/openwrt/pull/20632  (which also fixes firmware selector's metadata and re-enables sysupgrade's ability to reject bad images for that profile).

Which reminds me I need to get @hauke to backport it to 24.10 (and 25.12 if it's not already there).

efahl:
Yes, unneeded.

map-b:
This has become just bigger that it really is... Can we restart here?

Can we agree that replacing `,` by `_` in parallel that we do the real translation from `board_name` (usually dts compatible, comma ones) to `profile` (the underscore ones) is a hack/shorcut which usually matches the actual translation/mapping but is not the rigth thing to do at all? And that we should try to not guess the profile with any "sanitization" and do the real translation/mapping.

@aparcar the only feedback I got from you was this fully delegating, please, can I ask you for a real review? Should I apologize, should I introduce to myself before asking for review? It happens that we are in AI age and my slowly and carefully written, as I am trying to do my best, even for simple PR, are freely marked as IA slop and I am done?

Please, forget the specific gl-mt2500 case. I am just trying to make clearer how ASU handle the device profile identification process which has become really confusing with the "sanitization" here and there. I have looked in all actors: the server and clients. A proposal of "cleaning"(no functional change) in owut https://github.com/efahl/owut/pull/71

PR #1560, opened 2026-01-11
feat: add error logging

aparcar:
Basically @efahl work from #1522 with extra style changes

image

### DIFF
--- a/asu/routers/stats.py
+++ b/asu/routers/stats.py
@@ -1,8 +1,9 @@
 from datetime import datetime as dt, timedelta, UTC
 
 from fastapi import APIRouter
+from fastapi.responses import PlainTextResponse
 
-from asu.util import get_redis_ts
+from asu.util import error_log, get_redis_ts
 
 router = APIRouter()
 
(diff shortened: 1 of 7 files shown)


asu/util.py
```diff
@@ -575,3 +578,148 @@ def reload_profiles(app: FastAPI, version: str, target: str) -> bool:
...
+
+        Args:
+            build_request: The BuildRequest that failed
+            error_message: Description of the error
+        """
+        timestamp = datetime.now(UTC).strftime("%Y-%m-%d %H:%M:%S")
+        # Sanitize error message: single line, limited length
+        clean_error = " ".join(error_message.split())[:200]
+        profile_info = (
+            f"{build_request.version}:{build_request.target}:{build_request.profile}"
```
map-b: Can we somehow make clearer than we are not receiving a profile in most cases but a compatible string (firmware-selector and ASU's clients which try to get/translate the profile before sending the build_request are the exception)?

asu/routers/stats.py
```diff
@@ -120,3 +121,16 @@ def sum_data(version, data):
...
+def get_build_errors(n: int = 100) -> str:
+    """Return a summary of recent build errors.
+
+    Args:
+        n: Maximum number of entries to return (default 100, max 500)
+
+    Returns:
+        Plain text summary of build errors
+    """
+    return error_log.get_summary(min(n, 500))
```
efahl: 500 seems awful small.  We get about 600 errors per day and I was hoping view various time spans with tooling that talks to the API unrestricted.
```
curl -s 'https://sysupgrade.staging.openwrt.org/api/v1/build-errors?n=6000' | awk 'NR>5 {print $3}' | sort | uniq -c | sort -nr
```

efahl (changes requested):

efahl:
> `19 500 Server Error: Internal Server Error (no container with ID [job-id] found in database: no such container)`

Oh, now I remember this one.  That's how I found the extra `kill` and fixed with https://github.com/openwrt/asu/commit/ef20295d7d957bba8926aa900d36474d49417f9b

PR #1569, opened 2026-02-03
build: report package errors with a more detail

efahl:
Preliminary error statistics show that over 80% of all build errors are due to package errors. So instead of just reporting "Impossible package selection", attempt to show whether packages are missing or conflicting and list all packages named in the stderr from the build.

Packaging errors might arise from various causes:

 1) User specifies conflicting packages: nftables-json and nftables-nojson

 2) User specifies or has installed a package not from the OpenWrt feeds

 3) The upstream build of a package has failed, so it's appearing to be missing

This new log information will give us some insight as to whether these errors are caused by the user, the ASU client or upstream builds, so we know where to focus mitigations.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -18,6 +18,7 @@ from asu.util import (
     add_timestamp,
     add_build_event,
     check_manifest,
+    check_package_errors,
     diff_packages,
     error_log,
     fingerprint_pubkey_usign,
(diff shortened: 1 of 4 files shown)


efahl:
@aparcar 
I completely missed a case in opkg (`check_data_file_clashes`) and there's a bug in the apk conflicts regex, so part 2 incoming soon...

PR #1576, opened 2026-02-06
refactor: externalize package changes and branches to auto-reloading YAML

aparcar:
Move hardcoded package change rules and branch definitions into external YAML files (package_changes.yaml, branches.yaml) validated by JSON Schema and auto-reloaded on modification. No server restart required for updates.

Add CI step to validate YAML files against their schemas.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,9 @@ jobs:
       - name: Install dependencies
         run: uv sync --extra dev
 
+      - name: Validate YAML config files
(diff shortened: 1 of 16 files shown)


efahl:
Looking good, might take me a day or two for me to digest it all but I like what I'm seeing here.  I'll start running this locally and poking at it, see what I find.

aparcar:
Thanks, I'm still thinking about some details. Also I'd add an interface to visualize the changes

PR #1577, opened 2026-02-11
api: add versions to 'latest' API result

efahl:
In order to present a valid list of versions, downstream clients are currently parsing the upstream '.versions.json' and synthesizing (or not) a '-SNAPSHOT' version.  Since we're doing this already inside ASU, we might as well make it available to clients.

This will allow us to rewrite Firmware Selector, which has a hard-coded regex to filter unsupported versions (which is currently wrong).  With this change OFS will rely completely on the ASU server to tell it what is available.

Likewise, this allows a cleanup of owut to remove its similar parsing of the branches and .versions files.

It has no impact on LuCI ASU app, as it only uses the 'latest' list from this API call.

----
Additional comments

Should this actually be a new API: `api/v1/versions` instead of piling it onto the existing `api/v1/latest`???  This was simple and all the clients are already grabbing `latest`, so it seems like a good fit, but the endpoint name isn't the greatest.

----
Here's that regex, this PR pretty much eliminates all that code turning it into `versions = get("api/v1/latest").versions` and filtering out the -SNAPSHOT versions if the `show_snapshot` setting is turned off.

https://github.com/openwrt/firmware-selector-openwrt-org/blob/ea04cec0c9a507db3dabbc97ca6b6db4a58258e8/www/index.js#L828

### DIFF
--- a/asu/main.py
+++ b/asu/main.py
@@ -127,8 +127,17 @@ def generate_latest():
 
 @app.get("/json/v1/latest.json")
 def json_v1_latest():
+    """Returns two lists:
+
+    1) A list of the latest releases on each branch that is still
+       under support, including any upcoming RC versions.  Sorted by
+       release branch, with newest first.
+
+    2) A list of all available versions (both releases and snapshot),
+       sorted newest first.
+    """
     latest = generate_latest()
-    return {"latest": latest}
+    return {"latest": latest, "versions": app.versions}
 
 
 def generate_branches():
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -264,6 +264,13 @@ def test_api_latest_default(client):
     response = client.get("/api/v1/latest", follow_redirects=False)
     assert response.status_code == 301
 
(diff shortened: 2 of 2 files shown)


efahl:
And some more comments:

We're already using `settings.enabled` appropriately, but I think some branches are out of date.

I'd like for the `reload_versions` code to only supply a `xxx-SNAPSHOT` for those branches that actually have them, so...

The `settings.snapshot` attribute is currently unused in the code, it's simply reported to clients in the overview.  Can we make it mean "this branch has a -SNAPSHOT available", so as of this week's archive cleanups it would be false for 23.05, but true for 24.10 and 25.12?  Or should we have another attribute `snapshots_available` to do that?  

I'm looking at line 24 in the asu_schema.json of your refactoring PR.
```
"description": "Whether this is a snapshot (rolling) branch. Defaults to false."
```
could be "Whether this branch has a rolling (snapshot) version..."

EDIT:
I just dug through all of OFS, owut and LuCI app, none of them use the `overview.branch[*].snapshot` setting anywhere...

PR #1556, opened 2025-12-31
Add missing packages

hauke:
* package_changes.py: 25.12 Add kmod-dsa-mv88e6xxx
    
    The mv88e6xxx dsa driver is not compiled into the kernel any more. Add
    the package to the targets needing it.
    
    This was changed in OpenWrt in:
    https://github.com/openwrt/openwrt/commit/eaa82118eadfd495f8512d55c01c1935b8b42c51

 * package_changes.py: 25.12 Add PHY firmware for lantiq
    
    The PHY firmware is not compiled into the kernel image any more, but
    packages as a separate package. Add this firmware to the generated
    images if they are missing.
    
    This was changed in OpenWrt in:
    https://github.com/openwrt/openwrt/commit/a18d95f35bd54ade908e8ec3158435859402552d

PR #1579, opened 2026-02-16
package_changes.py: 25.12 Add switch driver for lantiq

### LINKED ISSUE #21334: No Ethernet Ports on BT Home Hub 5A after Attended Upgrade from 24.10.5 -> 25.12.0-rc1
### Describe the bug

Carried out an attended upgrade from 24.10.5 to 25.12.0-rc1
After the upgrade DSL synced and WiFi worked. Ethernet ports all missing.
Checking system log no mention of GSWIP, checking packages kmod-dsa-gswip wasn't installed (installing and rebooting wasn't enough).
Installed the pre-built 25.12.0-RC1 and that works as expected, ethernet available and packages manually added.

Could just be my config that was broken. Thought I'd flag and in others report it maybe worth digging.

If you are having the same issue just doing a "normal" install without resetting all config appears to fix it.

### OpenWrt version

openwrt-25.12.0-rc1-be7f7df573d0-lantiq-xrx200-bt_homehub-v5a-squashfs-sysupgrade

### OpenWrt release

25.12.0-rc1

### OpenWrt target/subtarget

lantiq/xrx200

### Device

BT Home Hub 5A

### Image kind

Official downloaded image

### Steps to reproduce

Carried out an attended upgrade
Following reboot no ethernet ports were seen in LUCI etc. No link / DHCP etc over wired links.
System Logs showed no messages about GSWIP

Added package kmod_dsa_gswip and rebooted
Failed to load with a error code 12.

Installed package from Firmware selector and came up fine.

Suspect either an issue with attended upgrade or something in my setup.

### Actual behaviour

Following attended upgrade wired ethernet not working. Wifi and DSL worked as expected.

Using package from firemware selector upgarded as expected.

### Expected behaviour

required modules for ethernet to be installed.

### Additional info

Have worked around, more recording in case it is a wider problem than just me.
I don't currently have a spare router to test / re-test on so haven't repeatedly reproduced it.

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

map-b:
The Lantiq xRx200/xRx300/xRx330 switch driver is not compiled into the kernel image any more. Add this switch driver to the generated images if it is missing.

This was changed in OpenWrt in:
https://github.com/openwrt/openwrt/commit/62bf0287326dcfab4596d5f4cad77cd9e7f8f03b

Related: https://github.com/openwrt/asu/pull/1556
Fixes: https://github.com/openwrt/openwrt/issues/21334

### DIFF
--- a/asu/package_changes.py
+++ b/asu/package_changes.py
@@ -130,8 +130,10 @@ def apply_package_changes(build_request: BuildRequest):
                 "solidrun_clearfog-pro",
             }:
                 _add_if_missing("kmod-dsa-mv88e6xxx")
+        # Changes for https://github.com/openwrt/openwrt/commit/62bf0287326dcfab4596d5f4cad77cd9e7f8f03b
         # Changes for https://github.com/openwrt/openwrt/commit/a18d95f35bd54ade908e8ec3158435859402552d
         elif build_request.target == "lantiq/xrx200":
+            _add_if_missing("kmod-dsa-gswip")
             if build_request.profile in {
                 "arcadyan_arv7519rw22",
                 "arcadyan_vgv7510kw22-brn",
@@ -168,8 +170,10 @@ def apply_package_changes(build_request: BuildRequest):
             }:
                 _add_if_missing("xrx200-rev1.1-phy11g-firmware")
                 _add_if_missing("xrx200-rev1.2-phy11g-firmware")
+        # Changes for https://github.com/openwrt/openwrt/commit/62bf0287326dcfab4596d5f4cad77cd9e7f8f03b
         # Changes for https://github.com/openwrt/openwrt/commit/a18d95f35bd54ade908e8ec3158435859402552d
         elif build_request.target == "lantiq/xrx200_legacy":
+            _add_if_missing("kmod-dsa-gswip")
             if build_request.profile in {
                 "alphanetworks_asl56026",
                 "netgear_dm200",


map-b:
This would be more correct since it is a `target`'s default package although the actual code reflects better the dependency driver-firmware or this work without the firmware in some case? Anyway I changed to this since it is target-wise.

```diff
diff --git a/asu/package_changes.py b/asu/package_changes.py
index ca12ba8..905d784 100644
--- a/asu/package_changes.py
+++ b/asu/package_changes.py
@@ -130,8 +130,10 @@ def apply_package_changes(build_request: BuildRequest):
                 "solidrun_clearfog-pro",
             }:
                 _add_if_missing("kmod-dsa-mv88e6xxx")
+        # Changes for https://github.com/openwrt/openwrt/commit/62bf0287326dcfab4596d5f4cad77cd9e7f8f03b
         # Changes for https://github.com/openwrt/openwrt/commit/a18d95f35bd54ade908e8ec3158435859402552d
         elif build_request.target == "lantiq/xrx200":
+            _add_if_missing("kmod-dsa-gswip")
             if build_request.profile in {
                 "arcadyan_arv7519rw22",
                 "arcadyan_vgv7510kw22-brn",
@@ -168,8 +170,10 @@ def apply_package_changes(build_request: BuildRequest):
             }:
                 _add_if_missing("xrx200-rev1.1-phy11g-firmware")
                 _add_if_missing("xrx200-rev1.2-phy11g-firmware")
+        # Changes for https://github.com/openwrt/openwrt/commit/62bf0287326dcfab4596d5f4cad77cd9e7f8f03b
         # Changes for https://github.com/openwrt/openwrt/commit/a18d95f35bd54ade908e8ec3158435859402552d
         elif build_request.target == "lantiq/xrx200_legacy":
+            _add_if_missing("kmod-dsa-gswip")
             if build_request.profile in {
                 "alphanetworks_asl56026",
                 "netgear_dm200",

```

efahl:
Good catch, there are more reports in the rc4 thread starting at https://forum.openwrt.org/t/openwrt-25-12-0-rc4-release-candidate/245939/165

Looks good to me.

@hauke?

hauke (approved):
LGTM. Sorry I missed that change before.

Issue #1583, opened 2026-03-04
[Traverse Ten64] AttributeError: 'NoneType' object has no attribute 'group'


DanaGoyette:
When I try to use attendedsysupgrade or owut, I get a server-side failure that seems rather cryptic.

It seems to be trying to call [.group()](https://github.com/openwrt/asu/blob/298ddb46d4d80005fc0043a7bbda33937c110a89/asu/build.py#L243) on a regular expression that it assumes will have matches.

I don't see any way to specify what "profile" to build; in my case, I'd need `traverse_ten64_mtd`.

I initially tried `owut` to upgrade from 24.10.x to 25.12.x, but I got the same error, so I had to manually upgrade.  Starting a new run of `owut` from 25.10.x gives the same error.

```
owut - OpenWrt Upgrade Tool 2026.01.13~2526d84b-r1 (/usr/bin/owut)
ASU-Server     https://sysupgrade.openwrt.org
Upstream       https://downloads.openwrt.org
Target         layerscape/armv8_64b
Profile        ten64-mtd
Package-arch   aarch64_generic
Root-FS-type   squashfs
Sys-type       sysupgrade
Version-from   25.12.0 r32713-f919e7899d (kernel 6.12.71)
Version-to     25.12.0 r32713-f919e7899d (kernel 6.12.71)
Build-commit   https://git.openwrt.org/?p=openwrt/openwrt.git;a=shortlog;h=f919e7899d
Build-FS-type  squashfs
Build-at       2026-03-03T00:14:15Z (~47 hours ago)
Image-prefix   openwrt-25.12.0-layerscape-armv8_64b-traverse_ten64_mtd
Image-URL      https://downloads.openwrt.org/releases/25.12.0/targets/layerscape/armv8_64b
Image-file     openwrt-25.12.0-layerscape-armv8_64b-traverse_ten64_mtd-squashfs-sysupgrade.bin
Installed      201 packages
Top-level      151 packages
Default         36 packages
User-installed 116 packages (top-level only)

Package version changes:
  luci                           26.061.65855~27a9c0c                        26.062.65959~3aa6773
  luci-app-attendedsysupgrade    26.061.65855~27a9c0c                        26.062.65959~3aa6773
  luci-app-firewall              26.061.65855~27a9c0c                        26.062.65959~3aa6773
  luci-app-package-manager       26.061.65855~27a9c0c                        26.062.65959~3aa6773
[...]

PR #1584, opened 2026-03-06
build: check profile name before usage

efahl:
If the build request contains a device profile that is not contained in the imagebuilder's device list, the build crashes with an internal server error.  This happens when the DTS for the device is incorrect, see link for example.

Fix by checking before attempting to extract data from the imagebuilder output.

Link: https://github.com/openwrt/asu/issues/1583

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -234,6 +234,9 @@ def _build(build_request: BuildRequest, job=None):
     )
     log.debug(f"Default packages: {default_packages}")
 
+    if f"\n{build_request.profile}:\n" not in job.meta["stdout"]:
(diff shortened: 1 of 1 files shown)


map-b:
> This happens when the DTS for the device is incorrect, see link for example.

This happens when the requested dts compatible/profile is valid in the profiles.json but it does not match any profile in imagebuilder. i.e. When overwriting DEVICE_NAME and it does not match the image profile name (original DEVICE_NAME) `ten64-mtd` vs `traverse_ten64_mtd`, solved using BOARD_NAME which does not affect the generated profiles.json

PR #1593, opened 2026-03-26
util: work around broken device naming

efahl:
There are a number of broken device specifications that redundantly map various images to the same device names.  Handle most of these by ensuring that the actual device name maps to itself, rather than some arbitrary value in 'SUPPORTED_DEVICES' due to these incorrect device mappings in the profiles.json.

Fixes: https://github.com/openwrt/openwrt/issues/22541

### DIFF
--- a/asu/util.py
+++ b/asu/util.py
@@ -642,11 +642,24 @@ def reload_profiles(app: FastAPI, version: str, target: str) -> bool:
         settings.upstream_url + f"/{version_path}/targets/{target}/profiles.json"
     )
 
-    app.profiles[version][target] = {
+    profiles = {
         name.replace(",", "_"): profile
         for profile, data in response.json()["profiles"].items()
         for name in data.get("supported_devices", []) + [profile]
     }
+
+    # Eliminate any remapping to alternate supported profiles.  Must happen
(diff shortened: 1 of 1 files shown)


aparcar:
Can we actually fix this upstream?

efahl:
> Can we actually fix this upstream?

Theoretically, yes.  Practically, it's spread around very thinly so they are hard to find (people don't always report the issues) and crosses version boundaries, so getting the PRs approved is pretty painful.

If this PR had logging as mentioned in its comments, that would give use pointers as to what is broken...

map-b:
It's a common practice to allow "force-less" sysupgrading between compatible images. I cannot call these cases "broken specs" as everything was implemented coherently (including sysupgrade paths, platform checks) and manual sysupgrade works as intended the only thing that need some love is "external", in the ASU side.

We have enough info to do it correctly, selecting the profile candidate which has the board_name entry the first in the SUPPORTED_DEVICES list.

*Indeed this information(shared entries) could be really useful for a potential enhancement for offering any of the supported images. (I probably shouldn't have said this last but I'm starting to be desperate, it seems) :( this cannot be healthy...

PR #1599, opened 2026-04-03
template: add info about installing clients via apk

Bruxism:
Links to referenced info about EoL of OpenWrt 24.xx, and the default inclusion of `luci-app-attendedsysupgrade` in 25.12+ is also included.

### DIFF
--- a/asu/templates/overview.html
+++ b/asu/templates/overview.html
@@ -38,11 +38,10 @@
                         >
                     </p>
                     <p>
-                        <b>How to use this server:</b> There are at least three
-                        clients that can request a new image from the Sysupgrade
-                        Server:
+                        <b>How to use this server:</b> These are clients 
+						that can request a new image from the Sysupgrade Server:
                     </p>
-                    <ul>
+                    <ul>For OpenWrt 24.10.xx
                         <li>
                             <a
                                 href="https://github.com/openwrt/asu#luci-app"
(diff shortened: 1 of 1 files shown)


Bruxism:
My first ever pull request. Thank you for encouraging me, @aparcar, and thank you for pointing me in the right direction, @efahl.

asu/templates/overview.html
```diff
@@ -57,19 +56,43 @@ <h2>About the Sysupgrade Server</h2>
...
+                    <p>
+                        <b>NOTE:</b> Following the release of 
+						<a href="https://openwrt.org/releases/25.12/notes-25.12.0#general_changes"
+						target="_blank">
+						OpenWrt 25.12</a>,
+						OpenWrt has transitioned from the traditional
+						<code>opkg</code> package manager
+						to <code>apk</code> (Alpine Package Keeper) and includes <code>
+						luci-app-attendedsysupgrade</code> by default in fresh installs.
+						We encourage everyone to upgrade to OpenWrt 25.12 before September 2026.
```
efahl: This mention of a specific version and date seems unmaintainable, would require updates "frequently", I'd just leave that line out.

Bruxism: I was leery of it myself, so reading what you said settles it. I'll take it out.

asu/templates/overview.html
```diff
@@ -38,11 +38,10 @@ <h2>About the Sysupgrade Server</h2>
                         >
                     </p>
                     <p>
-                        <b>How to use this server:</b> There are at least three
-                        clients that can request a new image from the Sysupgrade
-                        Server:
+                        <b>How to use this server:</b> These are clients 
+						that can request a new image from the Sysupgrade Server:
```
Bruxism, line 42: This feels awkward, and I'd like to redo it.

asu/templates/overview.html
```diff
@@ -57,26 +56,59 @@ <h2>About the Sysupgrade Server</h2>
                             <a
                                 href="https://github.com/openwrt/asu#cli"
                                 target="_blank"
-                                >Command line/SSH (OpenWrt >23.xx) :</a
+                                >Command line/SSH:</a
                             >
                             <code>opkg install owut</code>
                         </li>
+					</ul>
+					<ul>For OpenWrt 25.12.xx
```
Bruxism, line 64: I was thinking of putting something like (owut-based) and (apk-based) at the end of line 44/64 or maybe underlining them or some color or some kind of half-line break.

asu/templates/overview.html
Bruxism, line 37: I thought about putting these sentences elsewhere or putting some line breaks so that they stand on their own instead of presenting as one long line.

Bruxism:
I have some ideas for more changes, and am open to putting more thought into writing them, but right now, I feel the `apk` part is important enough to leave those for later that those other changes should be in another commit altogether.

asu/templates/overview.html
```diff
@@ -57,19 +56,42 @@ <h2>About the Sysupgrade Server</h2>
...
+                            >
+                        </li>
+						<li>
                             <a
                                 href="https://github.com/openwrt/asu#cli"
                                 target="_blank"
                                 >Command line/SSH:</a
                             >
-                            <code>opkg install auc</code>
+                            <code>apk -U add owut</code>
```
aparcar, line 81: Are you sure the -U is needed? I never used it :)

Bruxism: It's not needed per se because `apk update` could be run instead leading up to `apk add owut`. I included it just to make installation a one-line which would be easier for folk that are copy-pasting commands as I figure that anyone that knows to `apk update` and then `apk add ${package}` wouldn't be relying on it anyway.

I picked up the idea of including it when I was looking over other documentation to model this one.

Are there established conventions regarding this? I was considering context and aesthetics of the page for the reader.

aparcar: From my experience apk just downloads it on its own, however that might changes recently

PR #1622, opened 2026-04-16
feat(cache): revalidate cache

Guerra24:
To prevent mismatches between indexes and packages.

Valid time for 200 responses has been reduced to 1 minute. Squid used 0 seconds but with that setting Nginx does not cache the response.


aparcar:
Thanks, merged it :)

PR #1639, opened 2026-05-18
feat: support ROOTFS_FILESYSTEM via new `filesystem` field

aparcar:
Adds an optional `filesystem` field to BuildRequest (squashfs, ext4, ubifs, erofs, jffs2, jffs2-nand, targz, cpiogz). When set, the value is passed to ImageBuilder as `ROOTFS_FILESYSTEM=`, restricting the build to a single rootfs filesystem instead of producing all enabled ones.

Reduces build time and disk usage, and works around cases where one filesystem fails while another would succeed. Requires the OpenWrt side from openwrt/openwrt#23425; ignored on older ImageBuilders.

The request hash already reserved a slot for this field, so requests without `filesystem` produce the same hash as before and existing cached builds keep matching.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -330,6 +330,12 @@ def _build(build_request: BuildRequest, job=None):
                 f"ROOTFS_PARTSIZE={build_request.rootfs_size_mb}"
             )
 
+        if build_request.filesystem:
+            log.debug("Restricting filesystem to %s", build_request.filesystem)
+            job.meta["build_cmd"].append(
+                f"ROOTFS_FILESYSTEM={build_request.filesystem}"
+            )
+
         log.debug("Build command: %s", job.meta["build_cmd"])
 
         job.meta["imagebuilder_status"] = "building_image"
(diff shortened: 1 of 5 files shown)


efahl:
@aparcar  You're too quick for me!

I've been running this on my local server for a couple hours, ~30 `owut` builds across below targets/devices; still waiting for the upstream imagebuilders to pick up the actual functionality for real testing.  My biggest concern was that something in one of the oddball devices would find an edge case, but all good so far...

Here's the test device list, if you saw any weird ones I should add, let me know.
```
ath79/generic:elecom,wrc-300ghbk2-i
ath79/generic:tplink,archer-c7-v4
lantiq/xrx200:bt,homehub-v5a
mediatek/filogic:zyxel,nwa50ax-pro
mediatek/mt7622:linksys,e8450-ubi
mvebu/cortexa9:linksys_wrt3200acm
qualcommax/ipq807x:netgear_wax218
ramips/mt7621:netgear,r6220
ramips/mt7621:ubnt_edgerouter-x
x86/64:generic
```

PR #1641, opened 2026-05-20
docs: add server sysupgrade.libremesh.org

a-gave:
Hi, LibreMesh has a new official server for asu https://sysupgrade.libremesh.org, added it to the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,9 @@ immediately without rebuilding.
 > Unofficial servers, may run modified ImageBuilder
 
 - [ImmortalWrt](https://sysupgrade.kyarucloud.moe)
-- [LibreMesh](https://sysupgrade.antennine.org) (only `stable` and `oldstable` OpenWrt versions)
+- [LibreMesh](https://sysupgrade.libremesh.org)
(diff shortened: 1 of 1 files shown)


efahl:
@a-gave Should I be replacing the antennine entry in the `owut` docs with this, or do they serve different audiences?

https://openwrt.org/docs/guide-user/installation/sysupgrade.owut#servers

a-gave:
> @a-gave Should I be replacing the antennine entry in the `owut` docs with this, or do they serve different audiences?
> 
> https://openwrt.org/docs/guide-user/installation/sysupgrade.owut#servers

Yes it should be replaced with the new official instance of libremesh.
Actually https://sysupgrade.antennine.org is a backup server with official asu installed with feeds from libremesh enabled, it runs on a vps with 4 vcpu cores 8GB ram.
There is a second instance with my fork in an homelab https://syupgrade-01.antennine.org, 4cores 8GB ram. We plan to keep at least one of them as backup, because it is very convenient to have a server that pulls imagebuilders and build without having a local build environment.

PR #1660, opened 2026-07-22
Fix User-Agent crash, improve API correctness and reduce redundant calls

BKPepe:
- Fixes: 500 on missing User-Agent and Return 404 for unknown profiles

- Replace deprecated `datetime.utcfromtimestamp()`

- Use `logging` instead of `print()`** in the double-slash redirect handler.

- Cache Redis connections** in `get_redis_client()` — previously every call
  to `get_queue()`, `add_timestamp()`, or `get_redis_ts()` opened a new TCP
  connection. A single `/build` POST could trigger 3+ connections.

### DIFF
--- a/asu/build.py
+++ b/asu/build.py
@@ -86,11 +86,14 @@ def _detect_apk_mode(container) -> bool:
     return rc == 0
 
 
-def inject_files(container, build_request, job=None):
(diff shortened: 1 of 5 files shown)


BKPepe:
@aparcar Can you please take a look at this?

aparcar:
Thanks

PR #1662, opened 2026-08-15
ci: ruff: add explicit config

efahl:
As of ruff version 0.16.0, the default behavior of 'ruff lint' has changed from doing a subset of checks to doing all checks, including those from flake8-bugbear, pyupgrade, Pylint, and flake8-simplify. This causes a huge amount of noise in a simple 'ruff check' run, so revert the checks to just those we've been using (i.e., the old default set).


aparcar:
Sounds good but shouldn't we also work on making all ruff checks happy?

efahl:
> Sounds good but shouldn't we also work on making all ruff checks happy?

That would be a massive undertaking, with mostly a lot of code churn for no benefit.  The "ALL" setting calls even the tiniest lint an error, just for fun try picking through the noise in just one file that was considered clean before ruff 0.16.0.

```bash
$ ruff check --select ALL asu/build.py
warning: `incorrect-blank-line-before-class` (D203) and `no-blank-line-before-class` (D211) are incompatible. Ignoring `incorrect-blank-line-before-class`.
warning: `multi-line-summary-first-line` (D212) and `multi-line-summary-second-line` (D213) are incompatible. Ignoring `multi-line-summary-second-line`.
D100 Missing docstring in public module
--> asu/build.py:1:1

I001 [*] Import block is un-sorted or un-formatted
  --> asu/build.py:1:1
   |
 1 | / import datetime
 2 | | import json

... blah blah blah for 900 lines ...

Found 68 errors.
[*] 20 fixable with the `--fix` option (6 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

